PR #1, opened 2025-09-02
Port from sandbox

modcrafter77:
Move all code from sandbox to mainline


modules/api_ingress/src/lib.rs
```diff
@@ -0,0 +1,737 @@
...
+    pub error: String,
+    /// HTTP status code
+    pub code: u16,
+    /// RFC3339 timestamp when the error occurred
+    pub timestamp: String,
+    /// Optional request ID for tracking
+    pub request_id: Option<String>,
+}
+
+impl ErrorResponse {
```
striped-zebra-dev, line 56: I'd rather  use builder here, plus I think we'd better have some `restkit` library for that.

modules/users_info/src/infra/storage/mapper.rs
```diff
@@ -0,0 +1,13 @@
+use crate::contract::model::User;
+use crate::infra::storage::entity::Model as UserEntity;
+
+/// Convert a database entity to a contract model
+pub fn entity_to_contract(entity: UserEntity) -> User {
```
striped-zebra-dev, line 5: Don't we want to use `Into` for such cases?

jackofshadows7: Yes. We can. It's a half-generated sample.

modcrafter77:
TODO:
 1. ~~Eliminate `panic!/unwrap()` from libraries (registry, ingress, db locks, operation builder). Use proper error types.~~
2. Stabilize OpenAPI schema deduplication — canonical names + conflict checks.
3. Ensure graceful HTTP shutdown with lifecycle integration.
4. Keep `api_ingress` thin (gateway + OpenAPI registry); move sample routes to `examples/`.
5. Enforce domain → contract → HTTP error mapping into `ProblemResponse`.
6. CI: add workflows (fmt, clippy, tests, deny, audit, sccache).
7. Add `rust-toolchain.toml`, `justfile/Makefile`, clean up docs (remove “latest updates” logs).

** Nice-to-haves: **

* Metrics/logs for router rebuilds.
* Trace propagation (`request_id`, domain fields).
* Feature-gate heavy deps/assets.

PR #6, opened 2025-09-09
WIP Settings Module

striped-zebra-dev:


modules/settings/src/infra/storage/entity.rs
```diff
@@ -0,0 +1,93 @@
...
+pub struct NewSettingsEntity {
+    pub user_id: Uuid,
+    pub tenant_id: Uuid,
+    pub theme: String,
+    pub language: String,
+}
+
+/// Data for updating an existing user entity
+pub struct UpdateSettingsEntity {
+    pub theme: Option<String>,
```
Artifizer, line 34: Does it mean we are going to copy and copy all the model fields in NewXXX, UpdateXXX, Model, DTO, and such?
Isn't it too verbose?

modules/settings/src/infra/storage/mapper.rs
```diff
@@ -0,0 +1,26 @@
+use crate::contract::model::Settings;
+use crate::infra::storage::entity::Model as SettingsEntity;
+
+/// Convert a database entity to a contract model (owned version)
+impl From<SettingsEntity> for Settings {
+    fn from(rhs: SettingsEntity) -> Self {
```
Artifizer, line 6: why `rhs`? just curious

modules/settings/src/domain/service.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl Service {
+    pub fn new(db: DatabaseConnection, config: ServiceConfig) -> Self {
+        Self { db, config }
+    }
+
+    pub async fn get_settings(
+        &self,
+        user_id: Uuid,
+        tenant_id: Uuid,
```
Artifizer, line 34: Can we have a context object (with user_id and tenant_id and request_id, etc), not separate values?

modules/settings/src/infra/storage/entity.rs
```diff
@@ -0,0 +1,93 @@
+use chrono::{DateTime, Utc};
+use sea_orm::entity::prelude::*;
+use sea_orm::sea_query::OnConflict;
+use sea_orm::{ColumnTrait, EntityTrait, IntoActiveModel, QueryFilter, QueryOrder, QuerySelect, Set, TryIntoModel};
+use uuid::Uuid;
+use log::info;
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel)]
+#[sea_orm(table_name = "settings")]
+pub struct Model {
```
Artifizer, line 10: can we have a convention for the `Model` name here? Like, what if have several such 'models' in single file?

modules/settings/src/infra/storage/entity.rs
```diff
@@ -0,0 +1,93 @@
...
+/// Find a user's settings by user and tenant id
+pub async fn find<TX>(tx: &TX, user_id: Uuid, tenant_id: Uuid) -> Result<Option<Model>, DbErr>
+where
+    TX: ConnectionTrait,
+{
+    Entity::find_by_id((user_id, tenant_id)).one(tx).await
+}
+
+/// Update user's settings
+pub async fn update<TX>(
```
Artifizer, line 49: what if we have a generic upsert (save) method that handles DB timeout/retries, and such

striped-zebra-dev:
The code is out dated.

PR #10, opened 2025-09-14
feat: implement new ModKit DB Handling model with per-module DB configuration

modcrafter77:
…ation

BREAKING CHANGES:
- Replace monolithic database config with typed per-module configuration system
- Environment variable expansion is implemented for DB passwords (`${VAR}` syntax)
- Update ModuleCtx API: db() returns references, add config<T>() and raw_config() methods

### Major Features Added:
- **New Typed Configuration System**:
  - `GlobalDatabaseConfig`: Server-based configuration with auto-provisioning
  - `DbConnConfig`: Reusable connection config for global servers and modules
  - `PoolCfg`: Connection pool configuration with human-readable durations
  - `ModuleEntry`: Structured module configuration with database and config sections

- **Per-Module Database Handling**:
  - `build_final_db_for_module()`: Core precedence resolution engine
  - `PerModuleDbFactory`: Dynamic database provisioning for modules
  - Four-layer precedence: Global DSN → Global fields → Module DSN → Module fields
  - SQLite path resolution (file/path/default) with cross-platform support

- **Enhanced Environment Integration**:
  - `expand_env_in_dsn()`: ${VARNAME} expansion in DSN strings
  - `resolve_password()`: Handles both literal passwords and ${VAR} references
  - `validate_dsn()`: DSN validation using dsn crate

- **ModKit Runtime Integration**:
  - Updated `ModuleCtx` API with new db(), config<T>(), raw_config() methods
  - Enhanced module registry with per-module DB factory support
  - Conditional compilation for runtime-dependent features


modcrafter77:
You can check rationale here: https://gist.github.com/jackofshadows7/4a0c7f9d14c750c81f21d79cf179af86

apps/hyperspot-server/src/db_factory.rs
```diff
@@ -0,0 +1,311 @@
+//! Per-module database factory wiring.
```
Artifizer, line 1: I'd suggest to move majority of the code from the app to a library, because it should be reusable

jackofshadows7: removed at all

PR #54, opened 2025-12-05
feat: replace serde_yaml with serde_saphyr, closes #36

### LINKED ISSUE #36: Migrate config parsers to toml crate from serde_yaml
serde_yaml is unsupported crate already. Let's migrate to toml crate (https://crates.io/crates/toml)

Aelnor:
I have tested a couple of sanity check scenarios like `make check` and `make quickstart`.

The only remaining YAML mentions are Docker Compose and OpenAPI stuff.


README.md
```diff
@@ -55,49 +55,48 @@ make deny       # License and dependency checks
...
+max_backups = 3
+max_size_mb = 1000
 
 # Per-module configurations moved under modules section
-modules:
-  api_ingress:
-    bind_addr: "127.0.0.1:8087"
-    enable_docs: true
-    cors_enabled: false
+[modules.api_ingress.config]
```
Bechma: this should be only `modules.api_ingress` to be consistent with what we had

scripts/coverage.py
```diff
@@ -16,7 +16,10 @@
 from urllib.request import urlopen
 from urllib.error import URLError, HTTPError
 
-import yaml
+try:
+    import tomllib  # Python 3.11+
+except ImportError:
+    import tomli as tomllib  # Python < 3.11
```
Bechma: we need to specify this dependency or force to use python 3.11+
I would say the latest

README.md
Bechma: Forgive me, probably I misread the line where config was specified

PR #66, opened 2025-12-09
Event Broker REST API Design

striped-zebra-dev:


modules/event_broker/README.md
```diff
@@ -0,0 +1,761 @@
...
+- **Cursor-based listing**: `GET /v1/events` with filtering and pagination
+- **Long-polling**: `GET /v1/events:poll` - wait up to 30s for new events
+- **Sequence-based replay**: Read from any historical sequence position
+- **Type filtering**: Filter by single or multiple event types
+
+### Data Management
+
+- **Sequential ordering**: Guaranteed ordering within the log
+- **Efficient storage**: Optimized for high-throughput writes and range reads
+- **Type registry via ABI**: Topics and event types registered through local Rust client (not REST)
```
hitengajjar: What is ABI?

modules/event_broker/README.md
```diff
@@ -0,0 +1,761 @@
...
+      },
+      "data": {
+        "type": "object",
+        "properties": {
+          "userId": {
+            "type": "string"
+          },
+          "email": {
+            "type": "string",
+            "format": "email"
```
hitengajjar: do we have `email` format in JSON Schema for `string` data type?

MikeFalcon77 (approved):

modules/event_broker/README.md
```diff
@@ -0,0 +1,761 @@
...
+
+### Event Type
+
+**Fields:**
+
+- `id` (string, GTS Identifier) - Unique Global Event Type Identifier
+- `description` (string) - Human-readable description
+- `allowedSubjecTypes` (array of strings, GTS Identifiers) - Allowed subject types
+- `subjectRefTemplate` (string) - Subject reference template // TODO: do we still need this?
+- `dataSchema` (object) - JSON Schema defining event payload structure
```
hitengajjar: need to provide the expected schema (as shown in example)

modules/event_broker/README.md
```diff
@@ -0,0 +1,756 @@
...
+Problem: Allowing smooth updates to consumer filters without losing position in the event stream.
+
+Solution: Consumers are identified by a unique `consumer_id` and can be created with specific filter criteria.
+When a consumer's filters need to be updated, a new consumer instance should be created with the desired filters.
+The old consumer can be deleted once the new one is active. This approach ensures that each consumer maintains
+its own position in the event stream without conflicts.
+
+## Event Ordering and Pagination
+
+The Event Broker uses **offset-based pagination** (not cursor-based) for event consumption.
```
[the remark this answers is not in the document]
striped-zebra-dev: Please read, the entire section is there:
```
## Event Ordering and Pagination

The Event Broker uses **offset-based pagination** (not cursor-based) for event consumption.
This design aligns with event streaming semantics used in systems like Kafka.

- **Offset = Sequence Number**: The `offset` parameter represents the last seen event sequence number
- **Deterministic Ordering**: Events are ordered by monotonically increasing `sequence` field
- **Replay Capability**: Consumers can replay from any historical sequence number
- **Consumer Progress Tracking**: Offset represents explicit consumer position in the event log
```

modules/event_broker/README.md
```diff
@@ -0,0 +1,756 @@
+# Event Broker Module
+
+## Overview
+
+The Event Broker is a high-performance event streaming module for HyperSpot. It provides a REST API for producing and
+consuming events with support for batch operations, cursor-based pagination, and long-polling. Events are typed using
```
hyphen-2025: see resolved comment below

modules/event_broker/README.md
```diff
@@ -0,0 +1,756 @@
...
+    "CEL": "has(data) && data.userId == 'user-12345'",
+    "createdAt": "2025-12-03T12:05:00.000Z",
+    "expiresAt": "2025-12-03T12:05:30.000Z"
+  }
+}
+```
+
+#### GET /v1/events
+
+List events with cursor-based pagination.
```
hyphen-2025: see resolved comment below

Artifizer (changes requested):

modules/event_broker/README.md
```diff
@@ -0,0 +1,756 @@
...
+│   ├── gateways/          # Client implementations
+│   │   └── local.rs       # Local client for ClientHub
+│   ├── module.rs          # Module registration
+│   └── config.rs          # Typed configuration
+└── README.md
+```
+
+### Key Design Principles
+
+- **Sequential ordering**: Events are ordered by monotonically increasing sequence numbers
```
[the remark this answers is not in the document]
striped-zebra-dev: Let's discuss this, I don't get what IDs are we talking about here.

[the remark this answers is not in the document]
striped-zebra-dev: Fixed

modules/event_broker/README.md
```diff
@@ -0,0 +1,821 @@
...
+### Event Consumption
+
+- **Cursor-based listing**: `GET /v1/events` with filtering and pagination
+- **Long-polling**: `GET /v1/events:poll` - wait up to 30s for new events
+- **Sequence-based replay**: Read from any historical sequence position
+- **Type filtering**: Filter by single or multiple event types
+
+### Data Management
+
+- **Sequential ordering**: Guaranteed ordering within the log
```
nonameffh: It's not fully clear - what level of ordering is guaranteed: 
- Within a topic
- Within a topic and tenant
- Within a topic and an optional partition/routing key (that may/should include tenant ID)? 

In Constructor we actively use Kafka with ordering within a partition key

modules/event_broker/README.md
```diff
@@ -0,0 +1,821 @@
...
+      "data"
+    ]
+  },
+  "createdAt": "2025-12-03T12:00:00.000Z"
+}
+```
+
+### Event
+
+**Fields:**
```
nonameffh: Do we support something like headers in events?

modules/event_broker/README.md
```diff
@@ -0,0 +1,821 @@
...
+GET /v1/events:poll?consumer_id=cdb75c04-19e7-42ad-9999-d23f9d7c3392&offset=100
+```
+
+---
+
+#### POST /v1/consumers
+
+Create a consumer with complex filters for long-polling. This endpoint is optional and can be used to manage consumer
+state for multiple instances.
+The consumer is a temporary resource that holds filter criteria for a `sessionTimeout` period.
```
nonameffh: Could you explain how temporary a consumer group should be? And what's the lifecycle? 
For example in Kafka consumer groups by default live 7 days after all consumer group instances are disconnected, when consumer group instances are alive - information about consumer group offsets remains kept.

In Kafka we can use different strategies when creating a consumer group: a specific offset, earliest (offset = min known offset in a topic) and latest (offset = max known offset in a topic). I would suggest to support the same model for consumer creation with reflecting it to a sequence/session

modules/event_broker/README.md
```diff
@@ -0,0 +1,756 @@
...
+- **Batch publishing**: `POST /v1/events:batch` (up to 100 events) atomic writes, per-event error reporting
+- **Automatic sequence assignment**: Server-assigned monotonic sequence numbers
+- **Idempotency**: Duplicate sequence detection and conflict resolution
+
+### Event Consumption
+
+- **Cursor-based listing**: `GET /v1/events` with filtering and pagination
+- **Long-polling**: `GET /v1/events:poll` - wait up to 30s for new events
+- **Sequence-based replay**: Read from any historical sequence position
+- **Type filtering**: Filter by single or multiple event types
```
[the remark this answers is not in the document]
nonameffh: For consumer group management I believe we also need to cover the model of offset management (offset change when we have alive consumer event polling sessions may produce unexpected results)

nonameffh (changes requested):
From my perspective we also need to document the supported model of horizontal scaling (including how the sequence generation is implemented for multiple instances),  partitioning, replication for file-based streaming (is it supported?), data race prevention (could we use shared FS in multiple instances?)

PR #77, opened 2025-12-11
chore: add cast clippy rules

hyphen-2025:


libs/modkit-node-info/src/sysinfo_collector.rs
```diff
@@ -167,6 +168,7 @@ impl SysInfoCollector {
         }
     }
 
+    #[allow(clippy::cast_possible_truncation)]
```
Artifizer: can we better rewrite code like this?

Artifizer (approved):

PR #128, opened 2025-12-17
Implement api standards

Bechma:
from: https://gist.github.com/MikeFalcon77/c6c7eb2443f4cacd409692b7ea64ee11
- [x] JSON field naming
- [x] Location header on resource creation
- [x] Rate limiting headers (add governor) HEADS UP: rate limiting by request, we need to rate limit it by (Method, Path, AND IP)
- [x] Field projection ($select)
- [ ] OpenAPI vendor extensions()
- [x] Timestamp precision
- [x] API versioning in paths DONE https://github.com/hypernetix/hyperspot/pull/124
- [x] Problem Details error field naming
- [ ] PUT vs PATCH semantics
- [x] X-Request-Id propagation fixed on https://github.com/hypernetix/hyperspot/pull/130
- [ ] OpenAPI Vendor Extensions - MISSING


Artifizer:
I believe we need to make throttling :
- global
- per tenant

both must be in place, first - not to kill DB or service, second - to isolate noisy tenant

Bechma:
> I believe we need to make throttling :
> 
> * global
> * per tenant
> 
> both must be in place, first - not to kill DB or service, second - to isolate noisy tenant

We need to define how the throttling will affect maybe:
- Global means throttling by its IP?
- Per tenant means by the user token?

I added the requested headers to make the API standard and backward compatible, but we are throttling per method|route and that's a global that don't apply per user.

MikeFalcon77:
> * Global means throttling by its IP?
> * Per tenant means by the user token?
I suggest to use well-known throttling model. I will send PM with a link to ADN.

Artifizer:
> > * Global means throttling by its IP?
> > * Per tenant means by the user token?
> >   I suggest to use well-known throttling model. I will send PM with a link to ADN.

I think we can make it as separate PR

PR #130, opened 2025-12-18
bugfix(api_ingress): middleware stack order is fixed, test is added

MikeFalcon77:


modules/api_ingress/src/lib.rs
```diff
@@ -235,81 +308,15 @@ impl ApiIngress {
...
-        ));
-
-        // 9. Per-route rate limiting & in-flight limits (after MIME validation, before auth)
-        let rate_map = middleware::rate_limit::RateLimiterMap::from_specs(&specs, &config);
-        router = router.layer(from_fn(
-            move |req: axum::extract::Request, next: axum::middleware::Next| {
-                let map = rate_map.clone();
-                middleware::rate_limit::rate_limit_middleware(map, req, next)
-            },
+        // 10) Request ID handling
```
Bechma: to increase clarity, can we enumerate them in reverse? so the last one should be 1

striped-zebra-dev (approved):

Bechma (approved):

PR #131, opened 2025-12-18
refactor(runtime):add system pre_init/post_init hooks and simplify infra module wiring with OnceLock

MikeFalcon77:


Artifizer:
Let's update NEW_MODULE.md and docs/MODKIT_UNIFIED_SYSTEM.md explaining phases, their purposes and contracts, e.g. what error means - the module will be excluded? or the whole start process will be failed? Later we'll need to explain how it works with OoP modules as well

Artifizer:
Let's also update example to illustrate how to define own callbacks

libs/modkit/src/runtime/host_runtime.rs
```diff
@@ -216,6 +230,39 @@ impl HostRuntime {
         Ok(())
     }
 
+    /// `POST_INIT` phase: optional hook after ALL modules completed `init()`.
+    ///
+    /// This provides a global barrier between initialization-time registration
+    /// and subsequent phases that may rely on a fully-populated runtime registry.
+    ///
+    /// System modules run first, followed by user modules, preserving topo order.
```
Artifizer: where is the logic for user modules?

Artifizer:
I think the whole start flow should be adjusted to:

1. For every system module:
- run all pre_init's
- run all migrations
- run all init()
- run all post_init()

2. Then, for every user module do the same:
- run all pre_init's
- run all migrations
- run all init()
- run all post_init()
...

Maybe even have more "layers" ...

It would allow system modules to be fully ready to serve user modules, we can make an analogy between OS start and App start on top of the OS. Or modules start within kernel or initd

Artifizer:
OK, all these changes are non trivial and so we have now this dedicated task:
https://github.com/hypernetix/hyperspot/issues/132

Issue #120, opened 2025-12-15
Middleware: to finalize security context format


Artifizer:
@MikeFalcon77

PR #134, opened 2025-12-19
Adjust Claims to a final access structure, #120,

striped-zebra-dev:


libs/modkit-auth/examples/dispatcher_usage.rs
```diff
@@ -32,7 +32,7 @@ impl ClaimsPlugin for DemoClaimsPlugin {
         let sub_value = raw
             .get("sub")
             .ok_or_else(|| ClaimsError::MissingClaim("sub".to_owned()))?;
-        let sub = parse_uuid_from_value(sub_value, "sub")?;
+        let subject = parse_uuid_from_value(sub_value, "sub")?;
```
Artifizer: can we please move all these string tokens to a list of const's in single place?

PR #141, opened 2025-12-21
Restructure and reimplement Dylint lints

nanoandrew4:
This PR restructures the existing lints and re-implements them to make proper use of the AST/HIL, instead of relying on raw string matches or regexes. The restructuring aims to follow dylints best practices and houses each lint in its own folder, complete with its dependencies and test cases.

One limitation from dylint is when tests need external dependencies to run, you must use `ui_test_examples`, which copies the test files into a temporary directory for running the tests, which does not respect the original structure of the `ui` folder. This means that for all the lints that rely on file paths, the tests would always fail, since even if you put the files in `ui/contract/your_test_file.rs`, when the tests are run the intermediate `contract` directory is removed, and the lints won't know what module they belong to. This has been worked around by adding support for a simulated directory, which is marked by a comment on the first line of the test file that looks as follows: `// simulated_dir=/hyperspot/modules/some_module/api/rest/`. When the code detects it is in a temporary directory and sees this comment, it will use the path specified in the comment instead of the real one, which will allow the lints to run in the tests. Open to suggestions on how to improve this.

Also added two jobs to the CI workflow, one for running the lint tests, and one for running the lints against all our crates to make sure the code complies with the lints.


Bechma (commented):
can you add also the vscode/windsurf directives to the rust-analyzer?

Bechma (approved):

dylint_lints/de01_contract_layer/de0101_no_serde_in_contract/ui/deserialize.rs
```diff
@@ -0,0 +1,21 @@
...
+#[allow(dead_code)]
+// Should trigger DE0101 - Serde in contract
+#[derive(Debug, Clone, Deserialize)]
+pub struct Order {
+    pub id: String,
+    pub total: f64,
+}
+
+#[allow(dead_code)]
+// Should trigger DE0101 - Serde in contract
```
nanoandrew4: These comments are now enforced by the unit tests in each lint. If we were to change `Should trigger` for `Should not trigger`, or if we delete the comment altogether, the unit tests would fail, since that struct expects to have a lint error raised on it based on its use of `Deserialize`. This allows us to regain some of the atomicity that was spoken of, and override the `.stderr` files with more confidence, since this will act as an additional layer of intent checking for the tests.

Bechma (approved):

PR #142, opened 2025-12-22
feat(openspec): add OpenSpec guideline for Hyperspot mono-repo

hitengajjar:
Add comprehensive OpenSpec integration guide defining module-centric namespace structure and workflow for AI-driven development. Includes directory structure conventions, parallel development process, DOs/DON'Ts, ModKit integration steps, and Settings Service as reference implementation example.


hyphen-2025:
I believe `openspec init` generates its own prompts and workflow instructions to make AI agents stick to standard. As far as I understand this guideline shouldn't replace them but to complement to adapt them to monorepo structure. Does it make sense to commit also original openspec instructions? 

Or is our approach just to mimic openspec structure?

Artifizer:
@hitengajjar , we are going to switch to FDD and also have own taxonomy of the documents (PRD, DESIGN, ADR, and such) and so openspec no longer seen as main SDD framework for us. So I propose just to close this PR

hitengajjar:
We are moving to FDD and won't rely on OpenSpec, hence no need for this PR.

PR #148, opened 2025-12-23
Improve odata and openapi spec

### LINKED ISSUE #144: REST API guideline
Continuation of https://github.com/hypernetix/hyperspot/issues/43

Specification requirements:

https://github.com/hypernetix/DNA/blob/main/REST/QUERYING.md
```yaml
x-odata-filter:
  allowedFields:
    created_at: [ge, gt, le, lt, eq]
    id: [eq, in]

x-odata-orderby:
  allowedFields:
    - created_at desc
    - created_at asc
    - id asc
```

Current: These vendor extensions are not present in the OpenAPI spec.

PUT vs PATCH semantics
Update endpoints use PUT while accepting partial payloads.

The user_info sample need to be updated.

Bechma:
added $select and x-pagination as vendor extension in openapi response
closes https://github.com/hypernetix/hyperspot/issues/144


examples/modkit/users_info/users_info/src/api/rest/routes.rs
```diff
@@ -56,12 +56,23 @@ pub fn register_routes(
...
         .handler(handlers::list_users)
-        .json_response_with_schema::<modkit_odata::Page<dto::UserDto>>(openapi, http::StatusCode::OK, "Paginated list of users")
-        .with_odata_filter_doc("OData v4 filter. Examples: `email eq 'test@example.com'`, `contains(email,'@acme.com')`")
-        .query_param("$orderby", false, "OData orderby clause. Example: 'created_at desc, id desc'")
+        .json_response_with_schema::<modkit_odata::Page<dto::UserDto>>(
+            openapi,
+            http::StatusCode::OK,
+            "Paginated list of users",
+        )
+        .with_odata_filter()
```
MikeFalcon77: I think we need some single with_odata() with parameters instead of bunch of .with_odata_XXX. But I suggest to create an issue for this future improvement.

PR #149, opened 2025-12-24
update assets in example instead of build

### LINKED ISSUE #67: Fix regenerating modules/api_ingress/assets/elements/web-components.min.js
After make execution we typically see this:

git diff          
diff --git a/modules/api_ingress/assets/elements/web-components.min.js b/modules/api_ingress/assets/elements/web-components.min.js
index 1942d2c..a093e02 100644
--- a/modules/api_ingress/assets/elements/web-components.min.js
+++ b/modules/api_ingress/assets/elements/web-components.min.js
...

Bechma:
closes https://github.com/hypernetix/hyperspot/issues/67


Artifizer (approved):

modules/api_ingress/examples/update_assets.rs
MikeFalcon77: I suggest to move all examples under ./examples folder. Feel free to re-invent the structure of this folder

Bechma: I don't think having a common place for module examples is a good idea. Having examples of each module where you can execute a bunch of code related to that module is a better approach.

MikeFalcon77: Basically embedding was made for bare metal systems. These systems often doesn't have acces to Internet CDN

PR #153, opened 2025-12-25
fix(dylint): load all individual lint libraries in make dylint, added DE0110 lint to prohibit schema_for!() on GTS structs

Artifizer:
fix(dylint): load all individual lint libraries in make dylint
feat(dylint): add DE0110 lint to prohibit schema_for!() on GTS structs


Artifizer:
CI failed expectedly, that's exactly what these lints must catch

PR #157, opened 2025-12-26
feat: align GTS examples with v0.7.3 helpers and add DE0110 lint for GTS schemas

Artifizer:


libs/modkit/src/gts.rs
MikeFalcon77: Let's move this module to src/gts/plugin.rs. Because I believe we will have another core GTS entities in modkit.

examples/plugin-modules/tenant_resolver/plugins/fabrikam_tr_plugin/src/module.rs
```diff
@@ -40,31 +46,23 @@ impl Module for FabrikamTrPlugin {
...
-            .register(&SecurityCtx::root_ctx(), vec![schema_json])
-            .await?;
 
-        // Then, register the instance
-        let instance = ThrPluginSpec {
+        // === INSTANCE REGISTRATION ===
+        // Register the plugin INSTANCE in types-registry.
+        // Note: The plugin SCHEMA is registered by the gateway module.
+        let registry = ctx.client_hub().get::<dyn TypesRegistryApi>()?;
+        let vendor_clone = cfg.vendor.clone();
```
MikeFalcon77: You don't need this clone() It's a redundant heap copy

libs/modkit/src/gts.rs
Artifizer: https://github.com/hypernetix/hyperspot/pull/159

PR #163, opened 2025-12-26
 refactor(plugin-modules):: correct OpenAPI schema and simplify plugi…

MikeFalcon77:
…n initialization

- Use OnceLock instead of ArcSwapOption for one-time service setup
- Decouple gateway from concrete tenant resolver plugins
- Fix OpenAPI response type to match $select projections
- Centralize CSV status parsing logic
- Simplify plugin instance selection logic
- Normalize auth-related domain error mapping

- one grpc flaky test is improved

### DIFF
--- a/examples/plugin-modules/tenant_resolver/tenant_resolver-gw/src/api/rest/dto.rs
+++ b/examples/plugin-modules/tenant_resolver/tenant_resolver-gw/src/api/rest/dto.rs
@@ -85,6 +85,33 @@ impl From<TenantFilterDto> for tenant_resolver_sdk::TenantFilter {
     }
 }
 
+// ============================================================================
+// Shared Helpers
+// ============================================================================
+
+/// Parses a comma-separated status string into a `Vec<TenantStatus>`.
+///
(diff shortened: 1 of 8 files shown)


examples/plugin-modules/tenant_resolver/tenant_resolver-gw/src/api/rest/dto.rs
```diff
@@ -85,6 +85,33 @@ impl From<TenantFilterDto> for tenant_resolver_sdk::TenantFilter {
...
+///
+/// Recognized values (case-insensitive): `ACTIVE`, `SOFT_DELETED`, `UNSPECIFIED`.
+/// Unknown values are ignored.
+fn parse_statuses_csv(statuses: Option<&str>) -> Vec<TenantStatus> {
+    statuses
+        .unwrap_or("")
+        .split(',')
+        .filter_map(|raw| {
+            let s = raw.trim();
+            if s.eq_ignore_ascii_case("ACTIVE") {
```
Artifizer: I would use GTS everywhere for every constant - then we can do amazing validation/renaming/switch tricks

Artifizer (approved):

PR #164, opened 2025-12-27
[WIP] doc: added proposal specs for model_provider, oagw and cred_store components

Artifizer:
This is some sort of requirements for future services development/generation as part of preparation for Chat development. Here we prepare foundation for secure communication with external Model Providers, such as OpenAI, Anthropic, etc. The Models, Chat, File Parser and local Search will come next

### DIFF
--- a/dylint_lints/de08_rest_api_conventions/de0802_use_odata_ext/ui/invalid_odata.rs
+++ b/dylint_lints/de08_rest_api_conventions/de0802_use_odata_ext/ui/invalid_odata.rs
@@ -27,48 +27,47 @@ fn dummy_handler() {}
 
 pub fn define_endpoints() {
     // Using query_param for $filter - should use with_odata_filter
-    // Should trigger DE0802 - use OData ext
     OperationBuilder::get("/users-info/v1/users")
+        // Should trigger DE0802 - use OData ext
         .query_param("$filter", false, "OData filter expression");
 
     // Using query_param for $orderby - should use with_odata_orderby
-    // Should trigger DE0802 - use OData ext
     OperationBuilder::get("/users-info/v1/users")
+        // Should trigger DE0802 - use OData ext
         .query_param("$orderby", false, "OData ordering");
 
     // Using query_param for $select - should use with_odata_select
-    // Should trigger DE0802 - use OData ext
     OperationBuilder::get("/users-info/v1/users")
+        // Should trigger DE0802 - use OData ext
         .query_param("$select", false, "OData field selection");
 
     // Using query_param_typed for $filter
-    // Should trigger DE0802 - use OData ext
     OperationBuilder::post("/users-info/v1/users")
+        // Should trigger DE0802 - use OData ext
         .query_param_typed("$filter", false, "OData filter", "string");
 
     // Using query_param for $top
-    // Should trigger DE0802 - use OData ext
     OperationBuilder::get("/users-info/v1/users")
+        // Should trigger DE0802 - use OData ext
         .query_param("$top", false, "Maximum number of results");
 
     // Using query_param for $skip
-    // Should trigger DE0802 - use OData ext
     OperationBuilder::get("/users-info/v1/users")
+        // Should trigger DE0802 - use OData ext
         .query_param("$skip", false, "Number of results to skip");
 
     // Using query_param for $count
-    // Should trigger DE0802 - use OData ext
     OperationBuilder::get("/users-info/v1/users")
+        // Should trigger DE0802 - use OData ext
         .query_param("$count", false, "Include total count");
 
     // Multiple OData params in chain
-    // Should trigger DE0802 - use OData ext
     OperationBuilder::get("/users-info/v1/users")
+        // Should trigger DE0802 - use OData ext
         .query_param("$filter", false, "Filter")
         // Should trigger DE0802 - use OData ext
         .query_param("$orderby", false, "Order")
         .handler(dummy_handler)
         .register();
 }
-
 fn main() {}
--- a/dylint_lints/de08_rest_api_conventions/de0802_use_odata_ext/ui/invalid_odata.stderr
+++ b/dylint_lints/de08_rest_api_conventions/de0802_use_odata_ext/ui/invalid_odata.stderr
@@ -62,15 +62,6 @@ LL |         .query_param("$count", false, "Include total count");
    = help: use .query_param_typed() with proper OData extractor instead
    = note: type-safe OData methods provide compile-time validation and automatic OpenAPI schema generation
 
-error: use OperationBuilderODataExt instead of .query_param() for OData parameter `$filter` (DE0802)
-  --> $DIR/invalid_odata.rs:67:22
-   |
-LL |         .query_param("$filter", false, "Filter")
-   |                      ^^^^^^^^^
-   |
-   = help: use .with_odata_filter::<FilterFieldEnum>() instead
-   = note: type-safe OData methods provide compile-time validation and automatic OpenAPI schema generation
-
 error: use OperationBuilderODataExt instead of .query_param() for OData parameter `$orderby` (DE0802)
   --> $DIR/invalid_odata.rs:69:22
    |
@@ -80,4 +71,14 @@ LL |         .query_param("$orderby", false, "Order")
    = help: use .with_odata_orderby::<FilterFieldEnum>() instead
    = note: type-safe OData methods provide compile-time validation and automatic OpenAPI schema generation
 
+error: use OperationBuilderODataExt instead of .query_param() for OData parameter `$filter` (DE0802)
+  --> $DIR/invalid_odata.rs:67:22
+   |
+LL |         .query_param("$filter", false, "Filter")
+   |                      ^^^^^^^^^
+   |
+   = help: use .with_odata_filter::<FilterFieldEnum>() instead
+   = note: type-safe OData methods provide compile-time validation and automatic OpenAPI schema generation
+
 error: aborting due to 9 previous errors
+


openspec/changes/add-model-provider-cred-store-oagw/specs/oagw/spec.md
```diff
@@ -0,0 +1,610 @@
...
+Required metrics (using `metrics` crate):
+- `oagw.invocations` (counter, labels: tenant_id, route_id, status)
+- `oagw.request.duration` (histogram, labels: route_id, protocol)
+- `oagw.errors` (counter, labels: error_type, route_id)
+- `oagw.active_connections` (gauge, labels: route_id)
+- `oagw.circuit_breaker.opened` (counter, labels: link_id)
+- `oagw.circuit_breaker.state` (gauge, labels: link_id)
+- `oagw.bytes_sent` (counter, labels: tenant_id, route_id, link_id)
+- `oagw.bytes_received` (counter, labels: tenant_id, route_id, link_id)
+- `oagw.token_cache.hit_rate` (gauge)
```
hyphen-2025: Wouldn't it be better and more reliable to have counters with hits_total, misses_total and rate can be just derived from counter?

openspec/changes/add-model-provider-cred-store-oagw/specs/oagw/spec.md
```diff
@@ -0,0 +1,610 @@
...
+
+### Requirement: oagw health checks
+The system SHALL expose health check endpoints.
+
+REST endpoints:
+- `GET /oagw/v1/health` - Liveness probe
+- `GET /oagw/v1/ready` - Readiness probe (checks no circuit breakers open)
+- `GET /oagw/v1/routes/{routeId}/health` - Route-specific health check
+
+The health check endpoints access must be protected by role `oagw.health_check` assigned.
```
hyphen-2025: Does it mean if k8s tries to probe /health endpoint it needs to get auth token? For me it doesn't sound good.

openspec/changes/add-model-provider-cred-store-oagw/specs/cred-store/spec.md
```diff
@@ -0,0 +1,843 @@
...
+- `secret_type_id`
+
+#### Scenario: Envelope encryption
+- **WHEN** the plugin persists secret material
+- **THEN** the plugin encrypts the plaintext with a per-record DEK using an AEAD cipher
+- **AND** the plugin stores only ciphertext and encryption metadata
+- **AND** the plugin stores the DEK only in wrapped form (encrypted by KEK)
+
+### Requirement: cred_store SQLite-backed plugin
+The system SHALL support a `cred_store` plugin implementation that persists secrets in a local database (SQLite/SeaORM).
```
MikeFalcon77: probably for sqlite we should use this https://www.sqlite.org/see/doc/trunk/www/readme.wiki

openspec/changes/add-model-provider-cred-store-oagw/specs/cred-store/spec.md
```diff
@@ -0,0 +1,843 @@
+## ADDED Requirements
+
+### Requirement: cred_store module responsibility
+The system SHALL provide a `cred_store` gateway module with a **single active plugin (lowest priority)** responsible for tenant-scoped secret storage and retrieval.
```
MikeFalcon77: `lowest priority` is ambiguous and contradicts later statements where “lower wins”. Use one unambiguous rule: priority is an integer, lower value wins. I sugget to replace this phrase with 'single active plugin chosen by minimal priority value'.

openspec/changes/add-model-provider-cred-store-oagw/specs/cred-store/spec.md
```diff
@@ -0,0 +1,843 @@
...
+- Use scoped CRUD operations that automatically produce deny-by-default behavior for empty scopes.
+
+#### Scenario: DB-backed plugin denies cross-tenant reads
+- **GIVEN** a secret record exists for tenant A with secret id `S`
+- **WHEN** a caller from tenant B calls `CredStorePluginApi::get_secret_material(ctx, S, secret_type_id)`
+- **THEN** the plugin query is scoped to tenant B
+- **AND** the plugin returns a not-found or permission-denied error
+
+#### Cryptography (DB-backed plugins)
+DB-backed plugins SHALL encrypt secret material before persisting it.
```
MikeFalcon77: This implies plugins store the ciphertext as the source of truth. Right now you are designing two owners for the same data. Decide the ownership model:

Either gateway owns encryption and stores ciphertext, plugin is thin storage.

OR plugin owns encryption and stores ciphertext, gateway stores only metadata and plugin references.
Do not do both unless you want double-encryption and split-brain.

openspec/changes/add-model-provider-cred-store-oagw/specs/cred-store/spec.md
```diff
@@ -0,0 +1,843 @@
...
+
+#### Scenario: Audit log for failed access
+- **WHEN** user U attempts to read secret S but is denied (403)
+- **THEN** the gateway logs (client_id, 'read', secret_id, 'failure', 'access_denied')
+
+### Requirement: cred_store audit retention
+The system SHOULD support configurable audit log retention policies.
+
+Retention SHALL be configurable per tenant with defaults:
+- Default retention: 90 days
```
MikeFalcon77: Let's define a single policy model with explicit precedence: global default, per-tenant override, per-secret-type override and etc. Right now this reads like two unrelated rules.

openspec/changes/add-model-provider-cred-store-oagw/specs/cred-store/spec.md
```diff
@@ -0,0 +1,843 @@
...
+- `scope text null`
+- `created_at timestamptz not null`
+
+**Indexes:**
+- `(tenant_id, principal_id)` for fast role lookup
+
+NOTE: `cred_store_role_assignment` is currently out of implementation scope, it's assumed to be implemented by external access policy engine.
+
+#### Role definitions:
+- `gts.x.core.idp.role.v1~x.cred_store.secret.admin.v1`: full CRUD + key management + audit access
```
ast2074: should be separate roles for accessing metadata and the very secret

openspec/changes/add-model-provider-cred-store-oagw/specs/cred-store/spec.md
```diff
@@ -0,0 +1,843 @@
...
+The system must distinguish global per-tenant secret and per-user secret types.
+
+REST responses SHOULD avoid returning secret plaintext except in explicitly authorized retrieval modes.
+
+#### Scenario: REST list secrets
+- **WHEN** a client calls `GET /cred-store/v1/secrets`
+- **THEN** the response omits plaintext secret values
+
+### Requirement: cred_store plugin interface and selection
+The system SHALL define a plugin interface for secret storage backends.
```
ast2074: i'd suggest to consider only plugins for encryption, not for storage: 1) encryption by hs vault's transit mode, 2) custom own encryption impl

openspec/changes/add-model-provider-cred-store-oagw/specs/cred-store/spec.md
```diff
@@ -0,0 +1,843 @@
...
+
+NOTE: `cred_store_compliance_policy` support is out of scope of the first version.
+
+#### Scenario: EU-only data residency
+- **GIVEN** tenant T has data_residency = ['EU']
+- **WHEN** secret is created
+- **THEN** system ensures storage in EU-region database shard
+- **AND** prevents replication to non-EU regions
+
+### Requirement: cred_store GDPR right-to-erasure
```
ast2074: i'd suggest to generalize all compliance-related (gdpr, hippa, pci/dss etc.) requirements, they all are about few ideas actually, the main ones are audit (incl. retention) and rotating keks

openspec/changes/add-model-provider-cred-store-oagw/specs/oagw/spec.md
Artifizer: OK

openspec/changes/add-model-provider-cred-store-oagw/specs/oagw/spec.md
```diff
@@ -0,0 +1,610 @@
...
+- **WHEN** oagw invocation fails (timeout, 500 error)
+- **THEN** logs error details for troubleshooting
+- **AND** records error_message and status_code
+
+### Requirement: oagw observability
+The system SHALL expose Prometheus-compatible metrics.
+
+Required metrics (using `metrics` crate):
+- `oagw.invocations` (counter, labels: tenant_id, route_id, status)
+- `oagw.request.duration` (histogram, labels: route_id, protocol)
```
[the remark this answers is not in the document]
Artifizer: OK

openspec/changes/add-model-provider-cred-store-oagw/specs/model-provider/spec.md
```diff
@@ -0,0 +1,643 @@
...
+### Requirement: model_provider SLOs (Service Level Objectives)
+The system SHOULD target:
+- **Availability:** 99.9% (excluding planned maintenance and provider downtime)
+- **Latency:**
+  - p50: < 100ms for provider resolution
+  - p95: < 200ms for provider resolution
+  - p99: < 500ms for client resolution with cold cache
+- **Error rate:** < 0.1% (excluding provider errors and client errors)
+
+#### Scenario: SLO monitoring
```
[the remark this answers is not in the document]
Artifizer: Agree

openspec/changes/add-model-provider-cred-store-oagw/specs/cred-store/spec.md
Artifizer: OK

openspec/changes/add-model-provider-cred-store-oagw/specs/model-provider/spec.md
```diff
@@ -0,0 +1,633 @@
...
+The system SHALL define a Rust trait for provider plugins.
+
+```rust
+#[async_trait]
+pub trait ModelProviderPlugin: Send + Sync {
+    /// Plugin metadata
+    fn metadata(&self) -> &PluginMetadata;
+
+    /// Check if plugin can handle this provider configuration
+    fn can_handle(&self, vendor_gts_id: &str, kind_gts_id: &str, endpoint_type_gts_id: &str) -> bool;
```
MikeFalcon77: Looks like this call doesn't support all model selection criteries:

- capabilities
- features
- tenant requirements
- quotas
- priority

openspec/changes/add-model-provider-cred-store-oagw/specs/model-provider/spec.md
```diff
@@ -0,0 +1,633 @@
...
+
+#### Scenario: Revoke child access
+- **WHEN** parent tenant P sets `available_to_children = false` on provider
+- **THEN** child tenants lose access immediately
+- **AND** existing child usage is terminated gracefully
+
+### Requirement: model_provider audit logging
+The system SHALL maintain an audit log for all provider operations.
+
+#### Database table: model_provider_audit_log
```
MikeFalcon77: It will be a too hot table in case of high QPS. Probably we need to implement async audi logging via some message queue or use some time series (prometheus?)

openspec/changes/add-model-provider-cred-store-oagw/specs/model-provider/spec.md
MikeFalcon77: Moreover: what exactly is the ownership and granularity of plugins?

The spec currently does not clearly define what a plugin represents:
- a plugin per concrete LLM provider (e.g. OpenAI, Anthropic),
- a plugin per API/interface style (chat/completions, embeddings, streaming, etc.),
- or a plugin per concrete model family / endpoint variant.

This ambiguity affects:
- plugin selection logic (what `can_handle()` is really matching),
- capability/feature matching semantics,
- caching strategy (client cache key correctness),
- future extensibility (e.g. multiple APIs per vendor, mixed transports, local vs remote runtimes).

Suggestion:
- Explicitly define plugin ownership boundaries, for example:
  - plugins are responsible for a `(vendor + API style + transport)` combination, and models are data/config within providers, or
  - plugins are responsible for a `(vendor + model family)` and expose supported capabilities/endpoints declaratively.
- Document this as an invariant, not just an example, so both gateway logic and plugin authors have a single mental model.

guidelines/CODE_GEN_PIPELINE.md
```diff
@@ -0,0 +1,299 @@
...
+
+### 1.2 - Scaffold module layout
+
+1. Create module folder: `modules/{module_name}/`
+2. Create docs folder: `modules/{module_name}/docs/`
+3. Ensure you follow the canonical layout and SDK pattern from `guidelines/NEW_MODULE.md`.
+
+### 1.3 - Write intent design (`1_INTENT_DESIGN.md`)
+
+Use your favorite SDD (e.g. openspec-proposal) to generate `1_INTENT_DESIGN.md`.
```
hyphen-2025: IMO OpenSpec is very bad tool if we stick to "use your favourite tool approach". In such a case it's better to simulate OpenSpec ad-hoc without its injected prompts and AGENTS.md polluting context.

OpenSpec relies on ./openspec/spec to by always in-sync with current code. It's impossible unless every commits that adds new features or significantly changes behavior is done with OpenSpec flow.

guidelines/CODE_GEN_PIPELINE.md
```diff
@@ -0,0 +1,299 @@
...
+## Roles
+
+- **Human owner**: decides scope, accepts/declines changes, merges PRs.
+- **TOP-GRADE LLM**: creates architecture/code skeletons and higher-level reasoning.
+- **LOW-GRADE LLM**: fixes compilation/lints without changing intended behavior.
+
+## Non-negotiable project constraints
+
+- Follow:
+  - `guidelines/NEW_MODULE.md`
```
Bechma: I feel like we need to specify the "@" symbol to notate that we are talking about the content root, and not the relative path.

guidelines/CODE_GEN_PIPELINE.md
```diff
@@ -0,0 +1,299 @@
...
+- GTS schema and instances definitions
+```
+
+### 1.4 - Pre-PR self-review
+
+- Review the scope and missing invariants.
+- Review by other independent cloud models **ChatGPT**, **Gemini** and **Claude**
+
+```
+Review my 1_INTENT_DESIGN.md, criticize it, identify clear gaps regarding the industry
```
Bechma: criticize based on what?

modules/oagw/oagw-gw/src/domain/service.rs
```diff
@@ -0,0 +1,470 @@
...
+    /// Create a new service instance.
+    pub fn new(
+        route_repo: Arc<dyn RouteRepository>,
+        link_repo: Arc<dyn LinkRepository>,
+        secret_resolver: Arc<dyn SecretResolver>,
+        client_hub: Arc<ClientHub>,
+        config: ServiceConfig,
+    ) -> Self {
+        Self {
+            policy_engine: Arc::new(modkit_security::DummyPolicyEngine),
```
hyphen-2025: Construction of policy_engine in side new, violates DI principles. Dependency should be passed via constructor.

modules/oagw/oagw-gw/src/api/rest/dto.rs
```diff
@@ -0,0 +1,340 @@
...
+        }
+    }
+}
+
+// === Invocation DTOs ===
+
+/// HTTP method for invocation requests.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, ToSchema, PartialEq, Eq)]
+#[serde(rename_all = "UPPERCASE")]
+pub enum HttpMethodDto {
```
Bechma: Why? we already have axum::http or http package

modules/oagw/oagw-gw/src/api/rest/routes.rs
```diff
@@ -0,0 +1,375 @@
...
+        .operation_id("oagw.health")
+        .summary("Liveness probe")
+        .description("Returns healthy if the service is running")
+        .tag("oagw-health")
+        .public()
+        .handler(handlers::health)
+        .json_response_with_schema::<HealthResponse>(openapi, StatusCode::OK, "Service is healthy")
+        .register(router, openapi);
+
+    // GET /oagw/v1/ready - Readiness probe
```
Bechma: Why are we exposing two handlers that do the same? By convention there's the health endpoint to check the liveness of the services

modules/oagw/oagw-gw/src/api/rest/routes.rs
```diff
@@ -0,0 +1,375 @@
...
+        .require_auth(&Resource::Links, &Action::Read)
+        .require_license_features::<OagwLicense>([])
+        .query_param_typed(
+            "$top",
+            false,
+            "Maximum number of items to return",
+            "integer",
+        )
+        .query_param_typed("$skip", false, "Number of items to skip", "integer")
+        .query_param("$filter", false, "OData filter expression")
```
Bechma: this is not how we do it, please, refer to the example:
https://github.com/hypernetix/hyperspot/blob/main/examples/modkit/users_info/users_info/src/api/rest/routes.rs#L91

modules/oagw/oagw-gw/src/api/rest/routes.rs
```diff
@@ -0,0 +1,375 @@
...
+        .query_param("$orderby", false, "OData ordering expression")
```
Bechma: same situation, please, refer to example:
https://github.com/hypernetix/hyperspot/blob/main/examples/modkit/users_info/users_info/src/api/rest/routes.rs#L93

modules/oagw/oagw-gw/src/infra/storage/route_repo.rs
```diff
@@ -0,0 +1,225 @@
...
+            .one(self.conn.conn())
+            .await?;
+
+        let Some(route_model) = route_opt else {
+            return Ok(None);
+        };
+
+        // Load associated protocols
+        // Note: route_protocol is a join table without tenant_id - access is controlled via route
+        #[allow(clippy::disallowed_methods)]
```
hyphen-2025: According to clippy.toml we should 
`Use SecureSelect::all() after .secure().scope_with(scope)`

modules/oagw/oagw-gw/src/infra/storage/route_repo.rs
```diff
@@ -0,0 +1,225 @@
...
+                        route_id: id,
+                        protocol_gts_id,
+                    })
+                    .collect(),
+            ));
+        }
+
+        // Load current protocols
+        // Note: route_protocol is a join table without tenant_id - access is controlled via route
+        #[allow(clippy::disallowed_methods)]
```
hyphen-2025: Access should be controlled by AccessScope

modules/oagw/oagw-gw/src/infra/storage/route_repo.rs
```diff
@@ -0,0 +1,225 @@
...
+            .all(self.conn.conn())
+            .await?;
+
+        Ok(route_with_protocols(model, protocols))
+    }
+
+    async fn delete(&self, scope: &AccessScope, id: Uuid) -> anyhow::Result<bool> {
+        // Delete associated protocols first (cascade)
+        // Note: route_protocol is a join table without tenant_id - access is controlled via route
+        #[allow(clippy::disallowed_methods)]
```
hyphen-2025: Access should be controlled by AccessScope

modules/oagw/oagw-gw/src/lib.rs
```diff
@@ -0,0 +1,102 @@
+// Clippy allows for v1 implementation - to be tightened in v2
+#![allow(clippy::doc_markdown)] // Many technical terms without backticks
+#![allow(clippy::must_use_candidate)] // Will add systematically in v2
+#![allow(clippy::str_to_string)] // Prefer consistency over micro-optimization
```
Bechma: that's not correct, consistency means to use `to_owned` everywhere as this is how we approached in all the project.
We must attach to the project structure or drop the structure in favour of the definition of "consistency" for each person.

Bechma (changes requested):

guidelines/CODE_GEN_PIPELINE.md
```diff
@@ -0,0 +1,299 @@
...
+It is designed to be:
+
+- **LLM-actionable** (unambiguous inputs/outputs and prompt templates)
+- **Human-auditable** (review gates, checklists, clear phase boundaries)
+- **Project-aligned** (ModKit + DDD-light + SDK pattern, observability, security)
+
+## Roles
+
+- **Human owner**: decides scope, accepts/declines changes, merges PRs.
+- **TOP-GRADE LLM**: creates architecture/code skeletons and higher-level reasoning.
```
aviator5: It looks like this doc mixes 2 things: guideline for LLMs and development pipeline for humans. In my view it will confuse both.

For example, it introduces different LLM roles (top/low-grade) and recommends what kins of LLM should be used in different scenarious. It's useful for people, but what LLM should do with this info? At least for now when human is still responsible for choosing model for AI agent for processing a specific prompt.

I would suggest having 2 docs instead:
1. CODE_GEN_PIPELINE.md - Human-facing workflow with phases, gates, model recommendations, and references to prompts
2. CODE_GEN_PROMPTS.md - Clean LLM-facing prompt templates without role annotations or workflow context

This makes both documents more effective for their respective audiences.

guidelines/CODE_GEN_PIPELINE.md
aviator5: `CODE_GEN_PIPELINE.md` is currently in `guidelines/` dir but it's actually a human workflow for module developing and orchestrating LLMs, not guidelines that AI agents should follow. Suggest having a separate `workflows` dir and place this file there. This keeps `guidelines/` focused on norms/patterns that both humans and AI agents should follow (NEW_MODULE.md, SECURITY.md, etc.).

I guess we will implement tools that will automate pipelines from the `workflows/` dir and that will instruct AI agents to use guidelines from the `guidelines/` dir.

guidelines/CODE_GEN_PIPELINE.md
```diff
@@ -0,0 +1,299 @@
+# Module Code Generation Pipeline
+
+This document defines a repeatable, **human-controlled** pipeline for generating new modules with LLMs.
+It is designed to be:
+
+- **LLM-actionable** (unambiguous inputs/outputs and prompt templates)
```
aviator5: "LLM-actionable" confuses a bit - it's a pipeline for humans, not instructions for AI agents.

guidelines/CODE_GEN_PIPELINE.md
```diff
@@ -0,0 +1,299 @@
...
+Read: modules/{module_name}/docs/1_INTENT_DESIGN.md
+
+Generate:
+- modules/{module_name}/docs/2_INTENT_FACTS.md - commonly used facts across scenarios
+- modules/{module_name}/docs/3_INTENT_SCENARIOS.md - individual module scenarios - API, jobs if any, etc.
+
+Rules:
+- Facts must follow `FACTS_TAXONOMY.md`, you can use existing IDs or create new ones.
+- Every scenario step must reference a Fact ID (Fxx.yyy. if applicable).
+- Each scenario and each scenario step must be versioned with checkboxes:
```
aviator5: What **version** is here? Version of module/crate/public API?  I would suggest using the following terms:
1. Each **workflow** (please see my [previous comment](https://github.com/hypernetix/hyperspot/pull/164/files#r2651236615) for the details) consists of multiple **steps** (not **phases**).
2. `INTENT_DESIGN.md` describes architecture of the modules, its dependencies, implementations details, etc. All those things are divided into **phases** (phase - is a more familiar word here in my view). Naming: `<MODULE_PREFIX>-P<NUMBER>` (for better grepping and cross-referencing): `OAGW-P1`, `CREDSTORE-P42`, etc. I would suggest introducing the strict rule: every design aspect (authorization, tenancy isolation, licensing support, etc.) must have list of features divided into phases and some phase-to-phase migrations notes.
4. `INTENT_SCENARIOS.md` contains scenarios  aligned with **phases** from the `INTENT_DESIGN.md`. So, for example, `modules/oagw/docs/3_INTENT_SCENARIOS.md` may have two "[OAGW-P1] Timeout Configuration" and "[OAGW-P2] Timeout Configuration" which may be similar or completely different. Propose to **introduce ID for each scenario** (`<MODULE_PREFIX>-S<NUMBER>`): `OAGW-S1`, `CREDSTORE-S100500`, etc.
5. `INTENT_FACTS.md` contains facts (I believe we need to brainstorm to invent better name for this entity) aligned with **phases** from the `INTENT_DESIGN.md`. Approach is the same as for scenarios. Identifiers (`<MODULE_PREFIX>-F<NUMBER>`): `OAGW-F73`, etc.
6. Each module has **version** (version of the rust crate). I would suggest having `CHANGELOG.md` file that will
    - follow https://keepachangelog.com/en/1.0.0/ format
    - contain info what scenarios (human-readable phrase + ID) was added/fixed in which version.
Example:
```
  ## [0.2.0] - 2024-01-15
  ### Added
  - OAGW-S5: Timeout configuration support
  - OAGW-S12: Retry with exponential backoff
 ```

File: guidelines/CODE_GEN_PIPELINE.md (hunk 14 above)
Comment (aviator5) [reply guidelines/CODE_GEN_PIPELINE.md]: I'm not sure that we need to use any existing SDD tool for now at least for generating `DESIGN.md`, `FACTS.md`, `SCENARIOS.md`, etc.

File: modules/oagw/oagw-gw/src/config.rs [hunk 29]
```diff
@@ -0,0 +1,100 @@
...
+    /// Default connection timeout in milliseconds.
+    pub default_connection_timeout_ms: u64,
+    /// Default request timeout in milliseconds.
+    pub default_request_timeout_ms: u64,
+    /// Default idle timeout in milliseconds.
+    pub default_idle_timeout_ms: u64,
+
+    // === Rate Limiting (v2+) ===
+    /// Default rate limit per minute.
+    pub default_rate_limit_per_min: u32,
```
aviator5: Do we plan to support rate limits with granularity finer than “per minute” (e.g., 10 per sec)?
If yes, could we make the configuration more flexible-something like `10/s`, `100/m`, etc. - instead of hardcoding minute-based limits?

modules/oagw/oagw-gw/src/config.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// Timeout in seconds before half-open state.
+    pub circuit_breaker_timeout_sec: u32,
+    /// Number of successes in half-open to close circuit.
+    pub circuit_breaker_success_threshold: u32,
+
+    // === Audit Logging (v2+) ===
+    /// Audit guarantee level.
+    pub audit_guarantee: AuditGuarantee,
+    /// Maximum latency for audit logging in milliseconds.
+    pub max_audit_latency_ms: u64,
```
aviator5: `max_audit_latency_ms` -> `audit_max_latency_ms`?
for better consistency

modules/oagw/oagw-gw/src/module.rs
```diff
@@ -0,0 +1,141 @@
...
+        // Acquire DB with SecureConn for security-aware queries
+        let db = ctx.db_required()?;
+        let sec_conn = db.sea_secure();
+
+        // === SCHEMA REGISTRATION ===
+        // Gateway registers the plugin SCHEMA; plugins register their INSTANCES
+        let registry = ctx.client_hub().get::<dyn TypesRegistryApi>()?;
+
+        // Register OAGW plugin schema
+        let schema_str = OagwPluginSpecV1::gts_schema_with_refs_as_string();
```
MikeFalcon77: Accrosing to intent scenarios (5.1 Gateway Module Initialization), we should register:
gts.x.core.oagw.proto.v1~
gts.x.core.oagw.stream_proto.v1~
gts.x.core.oagw.auth_type.v1~
and Load well-known instances from gts/*.instances.json.

But I don't see it here.

modules/oagw/oagw-gw/src/config.rs
```diff
@@ -0,0 +1,100 @@
...
+
+/// OAGW module configuration.
+#[derive(Debug, Clone, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct OagwConfig {
+    // === Token Cache Settings (v3+) ===
+    /// Maximum capacity of the token cache.
+    pub token_cache_max_capacity: usize,
+    /// Maximum TTL for cached tokens in seconds.
+    pub token_cache_max_ttl_sec: u64,
```
MikeFalcon77: I think it should be std::time::Duration with humantime representation in yaml config.

modules/oagw/oagw-gw/src/domain/ports.rs
```diff
@@ -0,0 +1,49 @@
...
+use async_trait::async_trait;
+use modkit_security::SecurityContext;
+use oagw_sdk::Secret;
+use uuid::Uuid;
+
+use super::error::DomainError;
+
+/// Port for resolving secrets from cred_store.
+#[async_trait]
+pub trait SecretResolver: Send + Sync {
```
MikeFalcon77: Do we really need t oabstract secret resolver via ports pattern? I thought that we can call Credstore (whatever) client here.

Anyway ports in domain layer should be traits not concrete implementations. (https://github.com/hypernetix/hyperspot/blob/main/examples/modkit/users_info/users_info/src/domain/ports/audit.rs) 

The specific port implementation should be implemented in src/infra (https://github.com/hypernetix/hyperspot/blob/main/examples/modkit/users_info/users_info/src/infra/audit/http_audit_client.rs)

modules/oagw/oagw-sdk/src/api.rs
```diff
@@ -0,0 +1,189 @@
...
+///
+/// Plugins register themselves with a scoped client in `ClientHub`:
+/// ```ignore
+/// let scope = ClientScope::gts_id(&instance_id);
+/// hub.register_scoped::<dyn OagwPluginApi>(scope, plugin_impl);
+/// ```
+///
+/// The gateway resolves plugins at runtime based on protocol and auth type requirements.
+#[async_trait]
+pub trait OagwPluginApi: Send + Sync {
```
MikeFalcon77: I suggest to use XXXClient suffix. It's really client for some API but not API itself.

modules/oagw/oagw-gw/src/api/rest/routes.rs
```diff
@@ -0,0 +1,375 @@
...
+        .error_401(openapi)
+        .error_403(openapi)
+        .error_404(openapi)
+        .error_500(openapi)
+        .register(router, openapi);
+
+    // === Invocation Endpoint ===
+
+    // POST /oagw/v1/invoke - Invoke outbound API
+    router = OperationBuilder::post("/oagw/v1/invoke")
```
MikeFalcon77: I think we will have too many hops here. Let's imagine completion call:
Vendor client GET /v1/chat/completions -> API Gateway -> LLM Inference module (?) -> API Gateway -> OAGW -> OAGW Plugin -> real API (OpenAI API for ex.)

In case of Out Of Process modules the number of hops is increasing.

modules/oagw/oagw-gw/src/infra/storage/link_repo.rs
MikeFalcon77: I don't understand why do we need this repo. Links are a part of common data model of OAGW. Let's implemnt all these links operations in main repo.

MikeFalcon77 (changes requested):

guidelines/CODE_GEN_PIPELINE.md
Artifizer: Yes, this is a conceptual placeholder that will be refined later; however modern models already handle such loosely defined tasks quite well.

guidelines/CODE_GEN_PIPELINE.md
```diff
@@ -0,0 +1,299 @@
...
+/openspec-proposal
+
+Read: modules/{module_name}/docs/1_INTENT_DESIGN.md
+
+Generate:
+- modules/{module_name}/docs/2_INTENT_FACTS.md - commonly used facts across scenarios
+- modules/{module_name}/docs/3_INTENT_SCENARIOS.md - individual module scenarios - API, jobs if any, etc.
+
+Rules:
+- Facts must follow `FACTS_TAXONOMY.md`, you can use existing IDs or create new ones.
```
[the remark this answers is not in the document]
Artifizer: No, it needs to be added

modules/oagw/oagw-gw/src/api/rest/dto.rs
Artifizer: I believe it's because Dto must be a part of contract and dependency agnostic

modules/oagw/oagw-gw/src/api/rest/error.rs
```diff
@@ -0,0 +1,145 @@
+//! REST error mapping for OAGW.
+
+use http::StatusCode;
+use modkit::api::problem::Problem;
+
+use crate::domain::error::DomainError;
+
+/// Convert DomainError to Problem for REST responses.
+impl From<DomainError> for Problem {
```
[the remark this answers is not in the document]
Artifizer: @MikeFalcon77 what do you think?

modules/oagw/oagw-gw/src/api/rest/handlers.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    Ok(created_json(RouteDto::from(route), &uri, &id_str))
+}
+
+/// GET /oagw/v1/routes - List routes with OData pagination.
+#[tracing::instrument(skip(ctx, svc))]
+pub async fn list_routes(
+    Extension(ctx): Extension<SecurityContext>,
+    Extension(svc): Extension<Arc<Service>>,
+    OData(query): OData,
```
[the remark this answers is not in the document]
Artifizer: Yes, in the future versions

modules/oagw/oagw-gw/src/api/rest/routes.rs
Artifizer: In Kubernetes, readiness and liveness (“health”) probes serve different, very intentional purposes. Mixing them up causes real production issues, so this distinction matters.

Readiness: Can the pod handle traffic right now?
If it fails, Kubernetes stops sending traffic but does not restart the pod. This code will be implemented in v2 as it's mentioned in the "handlers::ready" handler comments

Liveness (health): Is the pod still alive?
If it fails, Kubernetes restarts the pod.

So we can use readiness for temporary or external issues; use liveness only when the process is broken and needs a restart.

modules/oagw/oagw-gw/src/domain/service.rs
Artifizer: Good catch, actually there are two problems - the one you mentioned and also there is DummyPolicyEngine is used, because this what was used in the example

openspec/changes/add-model-provider-cred-store-oagw/specs/model-provider/spec.md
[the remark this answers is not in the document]
Artifizer: Need to check where model management lives, should it be inside model provider or live separately?

openspec/changes/add-model-provider-cred-store-oagw/specs/model-provider/spec.md
Artifizer: Good catch. It should be about provider management, not requests - need to remove `invoke_model` and `resolve_client` from here

modules/oagw/oagw-gw/src/api/rest/routes.rs
Artifizer: Agree, created this:
https://github.com/hypernetix/hyperspot/issues/178

modules/oagw/oagw-gw/src/domain/ports.rs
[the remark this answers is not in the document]
Artifizer: I created this to investigate and fix guidelines/dylint: https://github.com/hypernetix/hyperspot/issues/179

modules/oagw/oagw-gw/src/api/rest/error.rs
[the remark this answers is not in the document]
Artifizer: Need a ticket

modules/oagw/oagw-gw/src/module.rs
```diff
@@ -0,0 +1,141 @@
...
+
+/// OAGW Gateway module.
+///
+/// This module provides:
+/// - Route and link configuration management
+/// - Plugin-based protocol and authentication handling
+/// - REST API for outbound API invocation
+#[modkit::module(
+    name = "oagw_gw",
+    deps = ["types_registry", "oagw_default_plugin"],
```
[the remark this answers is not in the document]
Artifizer: good catch:
https://github.com/hypernetix/hyperspot/issues/180

modules/oagw/oagw-gw/src/api/rest/error.rs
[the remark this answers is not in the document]
Artifizer: https://github.com/hypernetix/hyperspot/issues/191

modules/oagw/oagw-sdk/src/models.rs
```diff
@@ -0,0 +1,338 @@
...
+use chrono::{DateTime, Utc};
+use std::collections::HashMap;
+use std::pin::Pin;
+use uuid::Uuid;
+
+use crate::retry::RetryIntent;
+
+/// HTTP method for outbound requests.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum HttpMethod {
```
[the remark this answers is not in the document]
Artifizer: It was a part of Dto that typically is a duplication by it's nature as it's REST API contract

modules/oagw/oagw-gw/src/api/rest/dto.rs
```diff
@@ -0,0 +1,340 @@
...
+            supported_protocols: req.supported_protocols,
+        }
+    }
+}
+
+// === Link DTOs ===
+
+/// Response DTO for a link.
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "camelCase")]
```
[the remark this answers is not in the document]
Artifizer: good catch, I'll fix

PR #166, opened 2025-12-27
fix: update GTS library (gts-rust) to 0.7.5 and fix the e2e tests

claudedigon:
The 0.7.5 version prohibits using of 'gts.' prefix in JSON Schema $id values, it must start with "gts://gts."


Artifizer:
OK, Ubuntu tests fail as expected - known issue

PR #168, opened 2025-12-28
modkit-db: improve test coverage

hyphen-2025:
# Production Changes
## Database Transaction API change
- Changed transaction closure signatures in DbHandle:
with_pg_tx(), with_mysql_tx(), and with_sqlite_tx() 
- Removed separate Fut generic parameter, using for<'a> lifetime bounds instead

## OData Value Display Implementation

Added Display trait for modkit_odata::ast::Value
Enables string representation of OData values for debugging and logging
Supports all value types (Null, Bool, Number, String, etc.)

# Test changes

## Test Module Organization

- Restructured test modules in modkit-db/src/odata/:
- Disabled tests are rewritten
- Moved tests from monolithic tests.rs to organized module structure
- Sqlite in-memory tests are used for less brittle tests

## CodeCov report improvement
modkit-db coverage: from 60% to 81%
overall coverage: from 67% to 70%

https://app.codecov.io/github/hypernetix/hyperspot/commit/9966c7a415f892c305546b8da9bf28da23cae059/tree/libs?dropdown=coverage


libs/modkit-db/src/odata/tests/mod.rs
Bechma: To have everything better organized, would it be better to move all sqlite tests under a ./sqlite module where all tests inside don't require feature flags but just one in here?

libs/modkit-db/tests/transaction_sqlite.rs
```diff
@@ -0,0 +1,122 @@
+#![allow(clippy::unwrap_used, clippy::expect_used)]
+
+#[cfg(feature = "sqlite")]
```
Bechma: With the organization that I proposed, we can have /tests/odata/sqlite/* with all tests
And when we implement postgres, we can based the test structure of sqlite, to implement it for postgres.

MikeFalcon77 (approved):

Bechma (approved):

PR #170, opened 2025-12-28
fix(ci): Linux test job to disable debug info

hyphen-2025:
Problem: Tests with debug info consumed 17GB+ of disk storage and available space gets exhausted.

Solution: Run tests with flag disabling debug info - disk consumption ~5GB

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,8 +117,19 @@ jobs:
         run: cargo clippy --workspace --all-targets --all-features -- -D warnings -D clippy::perf
 
       # Unit tests (excluding integration tests)
+      - name: cargo clean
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -117,8 +117,19 @@ jobs:
         run: cargo clippy --workspace --all-targets --all-features -- -D warnings -D clippy::perf
 
       # Unit tests (excluding integration tests)
+      - name: cargo clean
+        if: runner.os == 'Linux'
+        run: cargo clean
+      - name: disk space
+        if: runner.os == 'Linux'
+        run: df -h && df -i
```
Bechma: why this df commands are here and below?
Looks like debugging info

PR #173, opened 2025-12-29
Remove humantime_serde crate in favor of stdlib duration parsing

### LINKED ISSUE #45: Drop humantime-serde in favour of chrono
https://github.com/jean-airoldie/humantime-serde/tree/master unmaintained
Better to use the most common chrono as we currently use it https://docs.rs/chrono/latest/chrono/

nanoandrew4:
Closes #45 

Removes `humantime-serde` since it is unmaintained and we can achieve almost the same functionality using the standard library. The only difference is now durations have to be integers in the yaml, which represent seconds, instead of allowing strings such as `5s`. The existing configs have been aligned with the new approach, so they can be viewed to see the difference.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2141,22 +2141,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "humantime"
-version = "2.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "135b12329e5e3ce057a9f972339ea52bc954fe1e9358ef27f95e89716fbc5424"
-
-[[package]]
-name = "humantime-serde"
-version = "1.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "57a3db5ea5923d99402c94e9feb261dc5ee9b4efa158b0315f788cf549cc200c"
-dependencies = [
- "humantime",
- "serde",
-]
-
-[[package]]
 name = "hyper"
 version = "1.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2933,7 +2917,6 @@ dependencies = [
  "dashmap",
  "dirs",
  "figment",
- "humantime-serde",
  "modkit-db-macros",
  "modkit-odata",
  "modkit-security",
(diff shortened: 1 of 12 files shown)


MikeFalcon77 (changes requested):
Removing humantime-serde because it is unmaintained is reasonable, but replacing human-readable durations with raw integer seconds is a regression in configuration UX/DX. These configs are authored by humans, and explicit units ("1s", "10m10s") are intentional, not a convenience.

We do not need humantime-serde to keep this behavior. A small local serde helper using humantime::parse_duration (which is stable and maintained) achieves the same result with minimal code. It requires ~200 LOC (reference: https://docs.rs/humantime-serde/latest/src/humantime_serde/lib.rs.html) in our codebase, so this change only removes functionality without providing a real benefit.

nanoandrew4:
The reason for dropping the duration strings was for simplicity and because from what I saw throughout the codebase and where the `humantime-serde` crate were used was for timeouts, which would tend towards double digit second counts, perhaps triple digit. IMO for those cases the DX is the same, since 88 seconds or 188 seconds is relatively easy to understand at a glance, you're unlikely to come across non-round second counts (I've never seen a config use something like 10m10s for a timeout, usually you would see 10m or maybe 10m30s), and throughout the codebase currently no timeout exceeds 60 seconds anyways from what I have seen.

Regardless, if we intend to have other durations in the future which are harder to understand at a glance in seconds, I see your point and would agree that the approach you suggest is probably more scalable to longer durations, which is where DX might fall off a cliff. I'm happy to close this PR and open another one with your suggested approach.

MikeFalcon77:
Thanks for the context, that’s fair.

Even if the current timeouts are short and easy to read in seconds, unitless integer durations tend to age poorly. In practice, we’ve repeatedly run into configs with integer values where it was unclear whether the unit was seconds, minutes, or hours, leading to guesswork or digging through code.

Human-readable durations avoid that ambiguity by making the unit explicit. We don’t need to keep humantime-serde for this. A small local serde helper using humantime::parse_duration gives us the same behavior with minimal complexity and no unmaintained dependency.

If you’re open to updating this PR to keep human-readable durations via a local helper, that would address both concerns without changing the overall direction.

nanoandrew4:
Closed in favor of #184

PR #184, opened 2025-12-30
Bring humantime-serde into modkit-utils to drop unmaintained dep

### LINKED ISSUE #45: Drop humantime-serde in favour of chrono

nanoandrew4:
Closes #45 

Implements alternative approach discussed in #173 where `humantime-serde` is implemented in a new `modkit-utils` crate, so we can keep the functionality the same while dropping the `humantime-serde` dependency, since it is unmaintained.

### DIFF
--- a/libs/modkit-utils/Cargo.toml
+++ b/libs/modkit-utils/Cargo.toml
@@ -15,7 +15,7 @@ serde = { workspace = true, optional = true }
 humantime = { workspace = true, optional = true }
 
 [dev-dependencies]
-serde_json = "1.0"
+serde_json = { workspace = true }
 
 [lints]
 workspace = true
--- a/libs/modkit-utils/Cargo.toml
+++ b/libs/modkit-utils/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "modkit-utils"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+authors.workspace = true
+repository.workspace = true
+description.workspace = true
+
+[features]
+humantime-serde = ["dep:humantime", "dep:serde"]
+
+[dependencies]
+serde = { workspace = true, optional = true }
+humantime = { workspace = true, optional = true }
+
+[dev-dependencies]
+serde_json = "1.0"
+version-sync = "0.8"
+
+[lints]
+workspace = true
--- a/libs/modkit-utils/Cargo.toml
+++ b/libs/modkit-utils/Cargo.toml
@@ -16,7 +16,6 @@ humantime = { workspace = true, optional = true }
 
 [dev-dependencies]
 serde_json = "1.0"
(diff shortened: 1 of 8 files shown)


libs/modkit-utils/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+#[cfg(feature = "humantime-serde")]
+pub mod humantime_serde;
+#[cfg(feature = "humantime-serde")]
+pub use humantime_serde::*;
```
Bechma: let's wrap this into a mod prelude {...}

libs/modkit-utils/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[features]
+humantime-serde = ["dep:humantime", "dep:serde"]
+
+[dependencies]
+serde = { workspace = true, optional = true }
+humantime = { workspace = true, optional = true }
+
+[dev-dependencies]
+serde_json = "1.0"
```
Bechma: use workspace version

Bechma (changes requested):

Bechma (approved):

libs/modkit-utils/src/humantime_serde.rs
```diff
@@ -0,0 +1,242 @@
...
+
+use humantime;
+use serde::{de, Deserializer, Serializer};
+
+/// Deserializes a `Duration` via the humantime crate.
+///
+/// This function can be used with `serde_derive`'s `with` and
+/// `deserialize_with` annotations.
+/// # Errors
+/// Returns a `humantime::Error` if string is not a valid Duration
```
MikeFalcon77: I think comment doesn't reflect reality. Looks like a formatting cannot fail only the serializer can return an error

MikeFalcon77 (approved):

libs/modkit-utils/src/humantime_serde.rs
```diff
@@ -0,0 +1,269 @@
...
+
+    #[test]
+    fn time() {
+        #[derive(Serialize, Deserialize)]
+        struct Foo {
+            #[serde(with = "super")]
+            duration: super::Duration,
+        }
+
+        let json = r#"{"duration": "10m 10s"}"#;
```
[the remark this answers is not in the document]
nanoandrew4: There seems to be a [crate](https://docs.rs/iso8601-duration/latest/iso8601_duration/) we could use, but `10m10s` is not ISO-8601 compliant, it would be `PT10M10S` from what I've seen.

Artifizer (approved):

PR #203, opened 2026-01-05
WIP: module scaffold tool and claude skill

hyphen-2025:
Main idea: deterministic scaffold tool will create a structure which is customized by agent.
This way we can enforce more rigid structure, prevent some hallucinations when agent tries to figure out how to use for example secure orm, and standardize module generation.

Bonus: less generation time and less token/premium request consumption

### DIFF
--- a/scripts/new-module-scaffold/new_module_acceptance.md
+++ b/scripts/new-module-scaffold/new_module_acceptance.md
@@ -1,0 +1,35 @@
+# New Module Acceptance Criteria
+
+Use this checklist to validate a newly scaffolded module before integration.
+
+General
+- [ ] Only files under `modules/<name>-sdk` and `modules/<name>` were created/modified by the tool.
+- [ ] No changes were made to root `Cargo.toml`, server Cargo.toml, or `apps/hyperspot-server/src/registered_modules.rs` (manual wiring instructions were printed instead).
+
+Build & Quality
+- [ ] `cargo check --workspace` succeeds.
+- [ ] `cargo fmt --all --check` passes.
+- [ ] `cargo clippy --workspace --all-targets -- -D warnings` passes for the new crates.
+
+API & Behavior
+- [ ] SDK defines `<PascalCase>Api`, `models::Health`, and `errors::<PascalCase>Error` (no `serde` derives in SDK types).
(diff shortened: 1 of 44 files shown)


scripts/new-module-scaffold/new_module_acceptance.md
```diff
@@ -0,0 +1,35 @@
+# New Module Acceptance Criteria
+
+Use this checklist to validate a newly scaffolded module before integration.
+
+General
+- [ ] Only files under `modules/<name>-sdk` and `modules/<name>` were created/modified by the tool.
+- [ ] No changes were made to root `Cargo.toml`, server Cargo.toml, or `apps/hyperspot-server/src/registered_modules.rs` (manual wiring instructions were printed instead).
+
+Build & Quality
```
Artifizer: I would just mention 'make all' or 'python ./scripts/ci.py all', because the list of actual checks may vary over time. For example we have `dylint` now

scripts/new-module-scaffold/new_module_acceptance.md
```diff
@@ -0,0 +1,35 @@
+# New Module Acceptance Criteria
+
+Use this checklist to validate a newly scaffolded module before integration.
+
+General
```
Artifizer: We need to mention explicitly that the module needs to be registered, otherwise it would not be even compiled

scripts/new-module-scaffold/new_module_acceptance.md
```diff
@@ -0,0 +1,35 @@
...
+- [ ] Only files under `modules/<name>-sdk` and `modules/<name>` were created/modified by the tool.
+- [ ] No changes were made to root `Cargo.toml`, server Cargo.toml, or `apps/hyperspot-server/src/registered_modules.rs` (manual wiring instructions were printed instead).
+
+Build & Quality
+- [ ] `cargo check --workspace` succeeds.
+- [ ] `cargo fmt --all --check` passes.
+- [ ] `cargo clippy --workspace --all-targets -- -D warnings` passes for the new crates.
+
+API & Behavior
+- [ ] SDK defines `<PascalCase>Api`, `models::Health`, and `errors::<PascalCase>Error` (no `serde` derives in SDK types).
```
Artifizer: Can you please double check that all the possible checks integrated into dylint checks?

Artifizer:
Let's close this one, as we now have a new design for CLI tool prepared by @Bechma

PR #215, opened 2026-01-08
Refactor/users info example revamp: transactions, OData streaming

MikeFalcon77:


examples/modkit/users_info/users_info/src/infra/storage/migrations/mod.rs
```diff
@@ -11,6 +13,8 @@ impl MigratorTrait for Migrator {
         vec![
             Box::new(initial_001::Migration),
             Box::new(add_tenant_support_002::Migration),
+            Box::new(add_relationships_003::Migration),
```
hyphen-2025: Btw is there any particular reason we have such unnatural naming convention for migrations?

It's not default one for seaorm (which is timestamp-base), and it's more difficult to sort files.

hyphen-2025 (commented):
Skimmed through changes. I like the approach. I will postpone my scaffold PR until this one is ready to merge.

examples/modkit/users_info/user_info-sdk/src/api.rs
```diff
@@ -1,51 +1,217 @@
...
+#[cfg(feature = "odata")]
+use crate::odata::{AddressSchema, CitySchema, LanguageSchema, UserSchema};
+
+/// Boxed stream type returned by streaming client facades.
+pub type UsersInfoStream<T> =
+    Pin<Box<dyn Stream<Item = Result<T, UsersInfoError>> + Send + 'static>>;
+
+/// Object-safe client for inter-module consumption (`ClientHub` registered).
+pub trait UsersInfoClient: Send + Sync {
+    fn users(&self) -> Box<dyn UsersStreamingClient>;
```
striped-zebra-dev: That is a very questionable API, what if we have 1 Mill users? will this open a transaction? in what isolation level? What kind of changes do we stream here?

examples/modkit/users_info/users_info/src/infra/storage/migrations/mod.rs
MikeFalcon77: I will rename these files

PR #230, opened 2026-01-13
Add lint to enforce humantime_serde usage for Duration in config structs

nanoandrew4:


nanoandrew4:
The lints are failing because the `modkit-transport-grpc` lib has a config struct that is missing the now mandatory annotations for duration serialization/deserialization. We can either add them despite them not being necessary for the fields in that config struct, since it doesn't seem like it is populated from a file, or we can weaken the lint to only look at files called `config.rs` or something like that, so that not all config structs need to have their Durations marked with the serde macro. I'm leaning towards the first, but am interested in collecting some feedback before making any changes.

nanoandrew4:
After speaking with @Bechma , we have decided to close this PR in favour of adding the recommendation to the guidelines. The issue is that no lint will reliably catch all durations that need to use the macro, and it would also force struct fields which don't need it to add it, which isn't helpful.

PR #236, opened 2026-01-14
feat(file_parser): add image format support for PNG, JPEG, WebP, and GIF

### LINKED ISSUE #221: file_parser: add image file support (png/jpg/webp/gif) and return base64 data
## Background

`file_parser` currently supports document formats (pdf, docx, html, txt) and produces a normalized IR (`ParsedDocument` + `ParsedBlock`).
For chat and multimodal LLM usage we also need to accept **image files** and return a representation that can be directly forwarded to vision-capable APIs.

This change intentionally does **not** introduce real image parsing or processing. Images are treated as opaque binary payloads and encoded as base64.

---

## Goal

Add support for standard image formats:

* PNG (`.png`, `image/png`)
* JPEG (`.jpg`, `.jpeg`, `image/jpeg`)
* WebP (`.webp`, `image/webp`)
* GIF (`.gif`, `image/gif`) – static only

For each uploaded image, return a `ParsedDocument` containing a single `ParsedBlock::Image` with a base64-encoded data URI.

---

### 1. New parser backend: `ImageParser`

Add a new backend in `src/domain/parsers/image_parser.rs` implementing `FileParserBackend`:

* `id() -> "image"`
* `supported_extensions() -> ["png", "jpg", "jpeg", "webp", "gif"]`
* `parse_local_path`: read file as raw bytes, determine MIME type, build IR
* `parse_bytes`: same logic, using `filename_hint` and/or provided content type

IR construction:

* `ParsedDocument.meta.content_type = image/<...>`
* `ParsedDocument.meta.original_filename` from `filename_hint` if available
* `blocks = [ ParsedBlock::Image { alt, title, src } ]`

`src` format:

```
data:<mime>;base64,<base64-encoded-bytes>
```

No decoding, resizing, or re-encoding is performed.

---

### 2. Content-Type and extension routing updates

Extend existing routing logic to recognize:

* `image/png` → `png`
* `image/jpeg` → `jpg` / `jpeg`
* `image/webp` → `webp`
* `image/gif` → `gif`

If upload content type is `application/octet-stream`, rely on `filename` extension as done today.

Unknown or unsupported binary formats should be rejected explicitly.

---

## REST API impact

No new endpoints are required.

Image support must work with existing endpoints:

* `POST /file-parser/v1/upload?filename=...`
* `POST /file-parser/v1/parse/url`

The returned response structure remains unchanged.

---

### E2E

Add E2E coverage using small binary fixtures:

* `tiny.png`
* `tiny.jpg`
* (optional) `tiny.webp`, `tiny.gif`

Test flow:

1. Upload image via `/file-parser/v1/upload`
2. Assert `document.meta.content_type == image/<type>`
3. Assert exactly one `Image` block is returned
4. Assert `src` starts with `data:image/<type>;base64,`
5. Decode base64 and verify byte-for-byte equality with the original file

Avoid snapshot-style comparisons due to large base64 payloads.

---

## Non-goals

* Image decoding or pixel inspection
* EXIF handling
* Resizing or normalization
* OCR or semantic analysis
* Animated GIF frame processing

This module should remain a transport-layer adapter only.

---

## Acceptance Criteria

* Image formats are detected correctly
* Base64 output preserves original bytes
* MIME type is propagated correctly
* Existing document parsing behavior remains unchanged
* All E2E tests pass

---

## Notes

If image transformation or inspection is required in the future, it should be implemented in a separate module to keep `file_parser` minimal and predictable.

bit8shift:
## Description
added image file support for png, jpg, webp and gif file formats, and return base64 data

## Type of Change
New feature (non-breaking change which adds functionality)

## Testing
Unit tests pass
Integration tests pass
Manual testing completed
New tests added for new functionality

## Checklist
Code follows project style guidelines
Self-review completed
No linting errors (`cargo clippy`)
Code is properly formatted (`cargo fmt`)
Tests pass (`cargo test`)

## Related Issues
Closes #221


Artifizer (approved):

modules/file_parser/src/domain/parsers/image_parser.rs
```diff
@@ -0,0 +1,164 @@
...
+
+impl ImageParser {
+    #[must_use]
+    pub fn new() -> Self {
+        Self
+    }
+
+    /// Determine MIME type from file extension
+    fn mime_type_from_extension(extension: &str) -> Option<&'static str> {
+        match extension.to_lowercase().as_str() {
```
MikeFalcon77: To eliminate redundant alloc here we can use eq_ignore_ascii_case  function:

fn mime_type_from_extension(extension: &str) -> Option<&'static str> {
    if extension.eq_ignore_ascii_case("png") { return Some("image/png"); }
    ....

modules/file_parser/src/domain/parsers/image_parser.rs
```diff
@@ -0,0 +1,164 @@
...
+/// Image parser that handles standard image formats
+///
+/// Treats images as opaque binary payloads and encodes them as base64 data URIs.
+/// No decoding, resizing, or re-encoding is performed.
+///
+/// Supported formats:
+/// - PNG (.png, image/png)
+/// - JPEG (.jpg, .jpeg, image/jpeg)
+/// - WebP (.webp, image/webp)
+/// - GIF (.gif, image/gif) – static only
```
MikeFalcon77: For LLMs we really should not send animated GIFs. But this code doesn't check this format. I suggest to leave some comment about it. We don't want to handle it via GIF parsing (for now) but we should document it.

modules/file_parser/src/domain/parsers/image_parser.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+}
+
+#[async_trait]
+impl FileParserBackend for ImageParser {
+    fn id(&self) -> &'static str {
+        "image"
+    }
+
+    fn supported_extensions(&self) -> &'static [&'static str] {
```
MikeFalcon77: please move it to const like: 
const EXT: &[&str] = &["png", "jpg", ...];
fn supported_extensions(&self) -> &'static [&'static str] { EXT }

MikeFalcon77 (changes requested):

modules/file_parser/src/domain/service.rs
```diff
@@ -280,6 +284,10 @@ impl FileParserService {
             "docx" => {
                 Some("application/vnd.openxmlformats-officedocument.wordprocessingml.document")
             }
+            "png" => Some("image/png"),
```
Bechma: Can we collect these static str into const so we can have easily mapped in references the:
Extension => MIME
MIME => Extension

modules/file_parser/src/domain/parsers/image_parser.rs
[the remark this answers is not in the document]
bit8shift: done :)

MikeFalcon77 (approved):

PR #241, opened 2026-01-14
Remove SecurityCtx from Calculator example

striped-zebra-dev:
## Summary by CodeRabbit

* **Chores**
  * Switched binary serialization from bincode to postcard and updated dependencies.
  * Renamed the security context type consistently across public APIs and SDKs; introduced Permission and SecurityContext exports.

* **Bug Fixes**
  * Removed an advisory ignore entry so the specified vulnerability is now subject to the configured advisory policy (now reported as a warning).

### DIFF
--- a/libs/modkit-security/src/bin_codec.rs
+++ b/libs/modkit-security/src/bin_codec.rs
@@ -1,6 +1,5 @@
-use crate::SecurityCtx;
-use bincode::config::{standard, Config};
-use bincode::error::{DecodeError, EncodeError};
+use crate::SecurityContext;
+use postcard::Error as PostcardError;
 use thiserror::Error;
 
 pub const SECCTX_BIN_VERSION: u8 = 1;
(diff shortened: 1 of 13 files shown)


libs/modkit-security/src/bin_codec.rs
```diff
@@ -1,19 +1,18 @@
...
+use postcard::Error as PostcardError;
 use thiserror::Error;
 
 pub const SECCTX_BIN_VERSION: u8 = 1;
 
 #[derive(Debug, Error)]
 pub enum SecCtxEncodeError {
     #[error("security context serialization failed: {0:?}")]
-    Bincode(EncodeError),
+    Postcard(PostcardError),
```
Bechma:
```suggestion
    Postcard(#[from] PostcardError),
```
With the from thiserror automatically implements the From trait

libs/modkit-security/src/bin_codec.rs
```diff
@@ -26,42 +25,42 @@ pub enum SecCtxDecodeError {
...
 
-    let cfg = secctx_config();
-    let payload = bincode::serde::encode_to_vec(ctx, cfg)?;
-    buf.extend_from_slice(&payload);
+    // Use a temporary buffer for postcard::to_slice to avoid requiring postcard's
+    // `alloc`/`std` features (postcard may be compiled with heapless output).
+    // 4 KiB should be enough for a typical SecurityContext; if it's larger, callers
+    // will get a serialization error and we can revisit.
+    let mut tmp = vec![0u8; 4 * 1024];
+    let payload_slice = postcard::to_slice(ctx, &mut tmp)?;
```
Bechma: you are allocating a Vec already, so you are going with the heap already as you are not using a slice of bytes.
Why not using `to_allocvec` so we don't need to care about heap/stack allocations? (always heap)

Bechma (changes requested):

Bechma (approved):

PR #262, opened 2026-01-16
Freeze stoplight frontend deps

ffedoroff:
Update .gitignore to include static frontend assets and remove web-components.min.js file

## Summary by CodeRabbit

* **Chores**
  * Pinned third‑party asset versions and replaced "latest" asset references with explicit versioned URLs.
  * Error messages now display the exact asset URL when downloads fail.
  * Added handling around asset download and local asset directory creation with clearer failure guidance.
  * No changes to runtime behavior, public APIs, or feature logic.


MikeFalcon77:
Discussed in chat. Basically we cannot delete this file from repo.

Bechma:
I think these files should be available during the build process to include them in the binary, but it makes little sense to actually have them in the git repo.

Bechma (approved):

MikeFalcon77:
> I think these files should be available during the build process to include them in the binary, but it makes little sense to actually have them in the git repo.

But what's your proposal?

MikeFalcon77:
> I think these files should be available during the build process to include them in the binary, but it makes little sense to actually have them in the git repo.

I agree but what storage options do we have?

PR #281, opened 2026-01-20
chore: improve scanner.rs

yoskini:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file exclusion matching so patterns are evaluated against both full file paths and individual file names, yielding more consistent and reliable filtering when excluding files from scans.

### DIFF
--- a/apps/gts-docs-validator/src/scanner.rs
+++ b/apps/gts-docs-validator/src/scanner.rs
@@ -35,12 +35,10 @@ fn gts_pattern() -> Regex {
 fn matches_exclude(path: &Path, exclude_patterns: &[Pattern]) -> bool {
     let path_str = path.to_string_lossy();
     for pattern in exclude_patterns {
-        if pattern.matches(&path_str) {
-            return true;
-        }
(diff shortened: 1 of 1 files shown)


apps/gts-docs-validator/src/scanner.rs
```diff
@@ -35,7 +35,11 @@ fn gts_pattern() -> Regex {
 fn matches_exclude(path: &Path, exclude_patterns: &[Pattern]) -> bool {
     let path_str = path.to_string_lossy();
     for pattern in exclude_patterns {
-        if pattern.matches(&path_str) {
+        if pattern.matches(&path_str)
+            || path
+                .file_name()
+                .is_some_and(|name| pattern.matches(&name.to_string_lossy()))
```
Bechma: just below we have this match, can you remove it?

Bechma (changes requested):

Bechma (approved):

MikeFalcon77 (approved):

Issue #279, opened 2026-01-20
oop-modules example is not working


binarycode:
**Steps to reproduce:**
- change `modules.grpc_hub.config.listen_addr` in `config/quickstart.yaml` to `127.0.0.1:50051`
- run `make oop-example` to execute hyperspot application
- run `cargo run -p calculator --features=oop_module` to run OoP module
- try to access calculator API:
```
curl -X POST http://localhost:8087/calculator-gateway/v1/calculator/add \
                 -H "Content-Type: application/json" \
                 -d '{"a": 5, "b": 3}' | jq

{
  "type": "about:blank",
  "title": "Internal Server Error",
  "status": 500,
  "detail": "Addition failed: internal error: CalculatorClient not available: client not found: type=dyn calculator_sdk::api::CalculatorClient",
  "instance": "",
  "code": "",
  "trace_id": "549755813889",
  "errors": null
}
```

Looks like there are multiple issues with `oop-modules` example:

1. **Calculator module is not actually OoP, it's compiled statically**

Here are the modules the application is compiled with (`apps/hyperspot-server/src/registered_modules.rs`):
```rust
#[cfg(feature = "oop-example")]
use calculator_gateway as _;

#[cfg(feature = "oop-example")]
use calculator as _;
```

Both `calculator_gateway` and `calculator` are statically compiled into the application, so there is no communication with OoP module. Also the `calculator_gateway` specifically declares dependency on `calculator` in `examples/oop-modules/calculator_gateway/calculator_gateway/src/module.rs`:
```rust
#[modkit::module(
    name = "calculator_gateway",
    capabilities = [rest],
    deps = ["calculator"]
)]
```

2. **`wire_client` is not used anywhere**

`wire_client` declared in `examples/oop-modules/calculator/calculator-sdk/src/wiring.rs` that should wrap gRPC connection to OoP `calculator` module is never called. This leads to `CalculatorClient not available` error, as this service is never registered in `ClientHub`.

I'm not sure what is the proper place to execute `wire_client`. My first thought was to put it into `CalculatorGateway#init` in `examples/oop-modules/calculator_gateway/calculator_gateway/src/module.rs`, but that seems to be too early - at this stage gRPC connection is not established yet.

PR #289, opened 2026-01-21
fix OoP modules example

### LINKED ISSUE #297: Unifying `deps` semantics for internal and OoP modules (soft OoP deps, no ghost modules)
ModKit uses `deps = ["..."]` to build a strict, topologically-sorted dependency graph of **in-process** modules. Today, if a dependency name is not a registered core module, registry build fails with an “unknown dependency” error. This makes it impossible to express dependencies on out-of-process (OoP) modules using the same `deps` list without introducing “ghost/stub” host modules.

ADR is attached: [adr-deps-oop-soft-deps.md](https://github.com/user-attachments/files/24810381/adr-deps-oop-soft-deps.md)

### LINKED ISSUE #298: Universal lazy OoP clients via ModKit (XXXClientProvider + codegen)
In ModKit, local (in-process) modules publish typed clients into `ClientHub` during `init()`, so consumers can call `hub.get::<dyn Trait>()` with no extra wiring. For OoP (out-of-process) modules and other remote dependencies (gRPC today, REST in the future), the current pattern often encourages consumers to do eager wiring (e.g. resolve endpoint + connect during module `init()`), which is fragile: OoP instances may be spawned after host start, and remote endpoints may be temporarily unavailable due to process start delays, scheduling, registration lag, or transient network issues.

We need a consistent and ergonomic architecture where:

- Consumer modules do **not** perform eager wiring (no `wire_client` in `init()`).
- The application can start even if an OoP dependency is temporarily unavailable.
- The absence of an OoP dependency affects only the specific operations that require it (e.g. HTTP 424 for those endpoints), not whole-module startup.
- The approach scales across many clients (gRPC now, REST later) without copy-pasting boilerplate.

ADR is attached:  [adr-universal-lazy-oop-grpc-clients.md](https://github.com/user-attachments/files/24810449/adr-universal-lazy-oop-grpc-clients.md)

binarycode:
Currently `oop-modules` example does not work (see #279).

There is no obvious place to run `wire_client` as proposed in documentation, so this PR attempts to fix the problem by introducing `LocalCalculatorClient` module that is expected to be compiled to the main application that exposes `CalculatorClient` interface. This module will lazily try to find connected `calculator` OoP module in `DirectoryClient` and call it if found. Thus this module effectively replaces `wire_client` function.

## Steps to reproduce the issue:
- run `make oop-example` to execute main application
- in a separate terminal run `MODKIT_DIRECTORY_ENDPOINT="unix:///tmp/hyperspot-grpc" cargo run -p calculator --features=oop_module` to execute OoP module
- in a separate terminal run `curl -X POST http://localhost:8087/calculator-gateway/v1/calculator/add -H "Content-Type: application/json" -d '{"a": 5, "b": 3}'`

**Response before fix:**
```json
{"type":"about:blank","title":"Internal Server Error","status":500,"detail":"Addition failed: internal error: CalculatorClient not available: client not found: type=dyn calculator_sdk::api::CalculatorClient","instance":"","code":"","trace_id":"274877906946","errors":null}
```

**Response after fix:**
```json
{"sum":8}
```

## Release Notes

* **New Features**
  * Added a local calculator module that automatically registers gRPC clients in the service hub with lazy initialization.
  * Calculator SDK now provides automatic client registration and service discovery integration.

* **Refactor**
  * Updated calculator SDK public API to expose the gRPC client directly and proto types for improved accessibility.
  * Module registration now uses automatic discovery instead of manual wiring configuration.


MikeFalcon77:
ghost modules for remote OoP modules looks not consistenet design solution.

I've created two issues with ADRs which will close a problems like this:

https://github.com/hypernetix/hyperspot/issues/298
https://github.com/hypernetix/hyperspot/issues/297

PR #300, opened 2026-01-23
fix: prevent unsafe PID conversion in process termination

### LINKED ISSUE #234: Propose proper fix for pull/160
Need to implement proper fix for https://github.com/hypernetix/hyperspot/pull/160 (Using DCO sign-off), and then decline this PR #160

bit8shift:
Closes #234 

## Summary

This PR fixes a dangerous PID conversion bug in Unix process termination logic.

## Problem

On Unix systems, converting a `u32` PID to `i32` using `unwrap_or(0)` could silently default to PID `0`.  
Sending a signal to PID `0` targets **all processes in the current process group**, which could unintentionally terminate unrelated processes during shutdown.

## Changes

- Replace `unwrap_or(0)` with explicit `let-else` conversion handling
- Log a warning when a PID exceeds `i32::MAX`
- Return `false` instead of signaling when conversion fails
- Improve documentation of failure modes and platform-specific behavior
- **Add tests covering PID conversion failure and termination edge cases**

## Behavior Change

- **Before:** Failed PID conversion could signal the entire process group
- **After:** Failed PID conversion logs a warning and no signal is sent

This makes shutdown behaviour safer and more predictable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved child process termination with stricter PID validation and handling; added debug logging when a termination signal is not sent
  * Unix path: better handling of large/invalid PIDs and signal delivery reporting
  * Windows path: explicit non-graceful termination behavior (no reliable graceful termination)
* **Documentation**
  * Expanded docs describing termination behavior on Unix and Windows
* **Tests**
  * Added tests covering termination behavior and edge cases

### DIFF
--- a/libs/modkit/src/backends/local.rs
+++ b/libs/modkit/src/backends/local.rs
@@ -27,6 +27,13 @@ const FORWARDER_DRAIN_TIMEOUT: Duration = Duration::from_millis(100);
 
 /// Send graceful termination signal to a child process.
 ///
+/// # Returns
+/// - `true` if signal was successfully sent
+/// - `false` if:
+///   - Process has no PID (already exited)
+///   - PID cannot be converted to i32 (extremely rare: PID > 2,147,483,647)
+///   - Signal delivery fails
+///
 /// On Unix: Sends SIGTERM which the process can handle for cleanup.
 /// On Windows: Returns false since there's no reliable graceful termination
 /// method for console applications.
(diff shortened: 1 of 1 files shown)


MikeFalcon77 (changes requested):

Bechma (changes requested):

MikeFalcon77:
test for macOS is failed. Please fix

Issue #214, opened 2026-01-08
Add new make/ci.py target to run hyperspot in docker compose with a real DB


Artifizer:
We need to have a new make targets to be able to run hyperspot in Docker compose in 2 flavors:
- with PostgreSQL DB
- with MariaDB 

Also there must be 2 other targets to run it in docker compose and to run e2e tests on top of it

Issue #259, opened 2026-01-16
Combined E2E Test Execution with Automatic Failure Ownership


MikeFalcon77:
The project contains an E2E test suite that validates the public API of a hyperspot.

Running the full E2E suite on every change is impractical, but running it too rarely delays detection of critical regressions.
Additionally, when E2E tests fail on `main`, ownership must be immediately clear to avoid stalled fixes.

### E2E Execution Strategy

We adopt a **combined execution strategy**:

1. **After merge to `main`**

   * Run a **reduced E2E smoke test suite**.
   * Purpose: fast detection of critical regressions that break the API or basic flows.
   * This run is considered blocking for the health of `main`.

2. **Nightly (scheduled)**

   * Run the **full E2E test suite**.
   * Purpose: deeper coverage without slowing down daily development.
   * This run is informational and does not automatically open new issues.

### Failure Ownership Model

* If the **smoke E2E suite fails on a push to `main`**:

  * A **GitHub Issue (bug)** is automatically created or updated.
  * The issue is **assigned to the author of the latest commit on `main`**.

* The latest commit is treated as the strongest ownership signal, regardless of merge strategy (merge commit or squash).

* Nightly failures do **not** automatically create issues to avoid noise.

  * They can be handled via manual triage or a long-lived tracking issue if needed.

### Operational Notes

* The smoke E2E suite must remain stable and fast.
* Flaky tests must be fixed or removed to avoid alert fatigue.
* Smoke and full E2E suites must be clearly separated at the test runner level.

---

### GitHub Actions Workflow

The following workflow implements this ADR.

Location: `.github/workflows/e2e.yml`

```yaml
name: e2e

on:
  push:
    branches: ["main"]
  schedule:
    - cron: "0 2 * * *" # nightly full E2E at 02:00 UTC
  workflow_dispatch:

permissions:
  contents: read
  issues: write

concurrency:
  group: e2e-${{ github.ref }}
  cancel-in-progress: false

jobs:
  e2e:
    runs-on: ubuntu-latest
    timeout-minutes: 45

    env:
      SERVICE_URL: http://127.0.0.1:8080
      HEALTH_PATH: /health

    steps:
      - name: Checkout
        uses: actions/checkout@v4

      - name: Set up Rust
        uses: dtolnay/rust-toolchain@stable

      - name: Set up Python
        uses: actions/setup-python@v5
        with:
          python-version: "3.11"

      - name: Install Python dependencies
        run: |
[...]

PR #320, opened 2026-01-27
ci: add new make targets to run hyperspot e2e tests with postgres and mariadb

ziqwerty:
#214

* Added new make targets - `make e2e-docker-pg` and `make e2e-docker-mysql`
* In docker compose, added new profiles, `postgres`, `mariadb` and `default` (sqlite), to be used in conjunction with e2e tests for targeted database environment.
* Fixed types_registry e2e tests to verify snake_case instead of CamelCase, complying with dylint's snake_case rules.
* Added a config var to allow insecure http for e2e tests.

## Release Notes

* **New Features**
  * Added multi-database support: PostgreSQL and MySQL/MariaDB alongside SQLite
  * New Docker Compose profiles enabling testing with different database backends
  * Build system now supports database driver selection during compilation

* **Chores**
  * Standardized API response field naming to use snake_case convention


modules/simple_user_settings/simple_user_settings/src/infra/storage/sea_orm_repo.rs
```diff
@@ -11,33 +12,97 @@ use crate::domain::repo::SettingsRepository;
...
+        ctx: &SecurityContext,
+        patch: SimpleUserSettingsPatch,
+    ) -> anyhow::Result<SimpleUserSettings> {
+        upsert_patch_impl(&self.db, ctx, patch, coalesce_patch_expr_pg_sqlite).await
     }
 }
 
 #[async_trait]
-impl SettingsRepository for SeaOrmSettingsRepository {
+impl SettingsRepository for SqliteSettingsRepository {
```
devjow: continuing from comment  https://github.com/hypernetix/hyperspot/pull/320/files#r2736091343

i think we can have smth like below:
```rust
#[async_trait]
impl SettingsRepository for DefaultSettingsRepository {
    async fn find_by_user(
        &self,
        ctx: &SecurityContext,
    ) -> anyhow::Result<Option<SimpleUserSettings>> {
        // implementation
    }

    async fn upsert_full(
        &self,
        ctx: &SecurityContext,
        theme: Option<String>,
        language: Option<String>,
    ) -> anyhow::Result<SimpleUserSettings> {
        // implementation
    }

    async fn upsert_patch(
        &self,
        ctx: &SecurityContext,
        patch: SimpleUserSettingsPatch,
    ) -> anyhow::Result<SimpleUserSettings> {
        // implementation
    }
}
```

what do you think?

Artifizer:
@ziqwerty , would you please review comments (incl. those coming from AI bots) and merge conflicts?

modules/simple_user_settings/simple_user_settings/src/infra/storage/sea_orm_repo.rs
ziqwerty: It seems there were refactoring made to DB repo to use orm, these changes are no longer needed. I have removed them.

modules/simple_user_settings/simple_user_settings/src/infra/storage/sea_orm_repo.rs
```diff
@@ -11,33 +12,97 @@ use crate::domain::repo::SettingsRepository;
...
+pub struct SqliteSettingsRepository {
     db: SecureConn,
 }
 
-impl SeaOrmSettingsRepository {
-    #[must_use]
-    pub fn new(db: SecureConn) -> Self {
-        Self { db }
+#[async_trait]
+impl SettingsRepository for PgSettingsRepository {
```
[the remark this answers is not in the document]
ziqwerty: Same https://github.com/cyberfabric/cyberfabric-core/pull/320#discussion_r2791312143

Bechma (changes requested):

libs/modkit/src/http/mod.rs
```diff
@@ -4,3 +4,24 @@
 //! modular web applications.
 
 pub mod sse;
+
+/// Conditionally enables insecure HTTP on an `HttpClient` builder in debug builds.
+#[macro_export]
+macro_rules! maybe_allow_insecure_http {
```
Bechma: Here there are still uses of debug_assertions, please, remove this macro and just use the value from the allow_insecure_http

Bechma (changes requested):

libs/modkit/src/http/mod.rs
MikeFalcon77: Discussed internally with Bruno. Let's leave this function as-is.
Rationale:
The concerns would be valid if this was about debug_assert!(). But this code does not use debug_assert!. It uses cfg(debug_assertions), which is a compile-time gate, not a runtime check.
 
Yes, cargo test, cargo run, and their -r variants use different profiles. That is expected and documented behavior. It is not “unreliable”. It is deterministic and fully controlled by Cargo profiles.

Different behavior between debug and release is normal in Rust and in systems programming in general. It is not a bug and not a design flaw by default. Rust itself does this: debug_assert! exists. overflow-checks differ by profile. panic = abort vs unwind and etc. 
 
Also we have no debug assertion logic leaking into prod. This is a compile-time restriction for a dangerous API.
The method is intentionally unavailable in release unless explicitly enabled.

Makefile
```diff
@@ -3,9 +3,6 @@ CI := 1
 OPENAPI_URL ?= http://127.0.0.1:8087/openapi.json
 OPENAPI_OUT ?= docs/api/api.json
 
-# E2E Docker args
```
Artifizer: Why it's removed? I see it's used below

Artifizer (changes requested):

Artifizer:
@ziqwerty , will it be possible to add CI as well? Right now I can't find it in .github/

Makefile
```diff
@@ -314,22 +313,30 @@ test-users-info-pg:
...
 
 # Run E2E tests locally
 e2e-local:
 	python3 scripts/ci.py e2e
 
-## Run E2E tests in Docker environment
+## Run E2E tests in Docker environment (default: sqlite)
 e2e-docker:
-	python3 scripts/ci.py e2e --docker $(E2E_ARGS)
+	python3 scripts/ci.py e2e --docker --cargo-build-args="--features $(DEFAULT_FEATURES)" $(E2E_ARGS)
```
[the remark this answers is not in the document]
ziqwerty: the default is db-sqlite already without explicitly defining it.

ziqwerty:
> @ziqwerty , will it be possible to add CI as well? Right now I can't find it in .github/

e2e has been added to github workflow, which runs nightly and will automatically create issue on failure. There is also e2e smoke for every push to `main`. In the description of issue #259, it mentions that running full e2e suite on every change might be impractical. Do u still prefer to have it in PR CI?

Bechma (approved):

MikeFalcon77 (changes requested):

testing/docker/init-postgres.sql
```diff
@@ -0,0 +1,46 @@
+-- Initialize databases for HyperSpot modules
```
Artifizer: Shall this code live inside testing/docker? What If I want to run Cyber Fabric on my dev machine or cloud server? How would I configure the database?

PR #469, opened 2026-02-05
fix ci build for dylint

Bechma:
## Summary by CodeRabbit

* **New Features**
  * Added a scheduled PR reviewer check that posts reviewer reminders and sends Telegram alerts for unreviewed PRs.

* **Bug Fixes / Improvements**
  * Standardized CI workflows: upgraded checkout actions, unified linting steps, replaced some builds with faster checks, and streamlined caching and job names.

* **Chores**
  * Simplified workspace and toolchain configurations and pinned select dependency versions.
  * Added CI-focused Make targets for tests and docs.

* **Governance**
  * Removed immediate PR-triggered reviewer notifications in favor of scheduled checks.


dylint_lints/Cargo.lock
```diff
@@ -835,178 +286,40 @@ dependencies = [
...
- "cfg-if",
- "crossbeam-utils",
- "hashbrown 0.14.5",
- "lock_api",
- "once_cell",
- "parking_lot_core",
-]
-
-[[package]]
-name = "de0101_no_serde_in_contract"
```
Artifizer: I believe we can not remove this rule de0101_no_serde_in_contract and need to replace it by it's equivalent.... Do we need an equivalent, btw or we do not? If we need and if you remove it now, we'll forget about it. Do we have a task for such replacement?

.github/workflows/ci.yml
```diff
@@ -450,78 +436,36 @@ jobs:
         if: env.SCCACHE_OK == '1'
         run: sccache --show-stats
 
-  dylint_tests:
```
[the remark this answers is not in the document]
Bechma: yes! both are included, but in the same job, it made no sense to run the lints if the tests for the lints were failing. This way we also save resources and speedup ci

dylint_lints/Cargo.lock
Bechma: I didn't remove it, I just disable it until we have a fix

Artifizer (approved):

PR #477, opened 2026-02-06
[WIP] [feat](settings_service): PRD, Design, and ADR documents as per FDD

hitengajjar:
# Settings Service - Multi-Tenant Configuration with GTS Schema & Compliance

Complete FDD documentation for Hyperspot's Settings Service module providing enterprise-grade configuration management with GTS schema validation, hierarchical tenant inheritance, and compliance features.

## 🔑 Top Key Features

1. **Multi-Tenant Hierarchical Inheritance** - Automatic value resolution cascading through 10+ levels of tenant hierarchy with configurable inheritance rules and barrier tenant support

2. **GTS Schema Validation** - Type-safe configuration with Global Type System schemas, JSON validation, and backward-compatible versioning

3. **Domain Scoped Settings Support** - Unified settings management across 2 domain types: Tenant, User

4. **Compliance Mode** - Setting locks, MFA-protected configurations, read-only enforcement, and full audit trail for regulated environments

5. **High Performance**

6. **Event-Driven Architecture** - Trait-based audit and notification events with configurable propagation modes (SELF, SUBROOT, NONE)

7. **Secure by Default** - Automatic tenant scoping, SecurityContext validation, role-based access control, and defense-in-depth architecture

8. Batch Operations - Efficient bulk updates and multi-tenant bulk get operations with partial success handling and detailed error reporting

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive design and PRD for the Settings Service covering architecture, API contracts, deployment, security, caching, value resolution/inheritance, and integration points.
  * Added 16 ADRs documenting decisions on schema versioning, database selection, caching, event integration, tenant hierarchy, API versioning, performance, registry, trait enforcement, defaults, compliance locks, soft deletion, batch transactions, and related patterns.


modules/system/settings_service/docs/adrs/0002-fdd-settings-service-adr-database-technology.md
```diff
@@ -0,0 +1,50 @@
+# ADR-0002: Database Technology Selection
```
Artifizer: I would remove this ADR because we are DB agnostic and there is no need to touch this topic in every PR unless we do want to become DB-specific by a reason

modules/system/settings_service/docs/adrs/0005-fdd-settings-service-adr-tenant-hierarchy-cache.md
```diff
@@ -0,0 +1,69 @@
...
+
+**Status**: Proposed
+
+**ADR ID**: `fdd-settings-service-adr-tenant-hierarchy-cache`
+
+## Context and Problem Statement
+
+⚠️ **IMPLEMENTATION NOTE**: This ADR is deferred for the initial implementation. The initial version will query the Tenant Management Module database directly for tenant hierarchy data. Event-driven cache synchronization will be implemented in a future phase as a performance optimization.
+
+The Settings Service needs to maintain a local cache of tenant hierarchy data for efficient inheritance resolution. We need to determine how to synchronize this cache with the Tenant Management Module and handle hierarchy changes in real-time.
```
Artifizer: Please double check with @aviator5 regarding tenant hierarchy

modules/system/settings_service/docs/adrs/0006-fdd-settings-service-adr-api-versioning.md
```diff
@@ -0,0 +1,49 @@
+# ADR-0006: API Versioning and Deprecation Policy
```
Artifizer: This ADR is not needed as it's not a per-module decision anymore, this is global enforced decision. The code will not compile if there is no major version provided, this is protected by dylint

modules/system/settings_service/docs/adrs/0007-fdd-settings-service-adr-performance-optimization.md
```diff
@@ -0,0 +1,49 @@
+# ADR-0007: Performance Optimization Approach
```
Artifizer: I'd merge it with 0003 (caching)

modules/system/settings_service/docs/adrs/0008-fdd-settings-service-adr-gts-registry-implementation.md
```diff
@@ -0,0 +1,50 @@
+# ADR-0008: GTS Type Registry Implementation
```
Artifizer: This ADR is not needed as Cyber Fabric has the only Type Registry service that is always accessed via native Rust SDK. This SDK will make all the needed gRPC/HTTP transport under the hood if need automatically

modules/system/settings_service/docs/adrs/0010-fdd-settings-service-adr-default-value-resolution.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Considered Options
+
+* **Option 1**: Schema-based defaults stored in GTS schema definition
+* **Option 2**: Configuration-based defaults in service configuration
+* **Option 3**: Database-stored defaults with version tracking
+
+## Decision Outcome
+
+Chosen option: "Option 1 - Schema-based defaults stored in GTS schema definition", because it ensures defaults are validated, versioned with the schema, and accessible without additional database queries or configuration lookups.
```
Artifizer: Well, as of TENANT settings - I believe we should have at least two strategies for the defaults if a values on given tenant is not set.

Strategy 1 - use schema defaults
Strategy 2 - check parents one by one, use the first parent with non-default values and use it. Otherwise use schema default values

modules/system/settings_service/docs/adrs/0016-fdd-settings-service-adr-database-schema-hierarchy.md
```diff
@@ -0,0 +1,146 @@
...
+
+1. **Get all ancestor settings for inheritance**:
+
+```sql
+SELECT * FROM setting_values 
+WHERE setting_type_id = ? 
+  AND tenant_path IN (
+    '/dc52a314-6d47-44f7-90e8-48d5219cbc61/',
+    '/dc52a314-6d47-44f7-90e8-48d5219cbc61/259a8820-ef53-40e8-b3a7-a0f78a076c06/',
+    '/dc52a314-6d47-44f7-90e8-48d5219cbc61/259a8820-ef53-40e8-b3a7-a0f78a076c06/2e7c1d4f-ecd0-4fcb-9f08-6e9d9f5d909e/'
```
Artifizer: I think this is bad idea because:
1. It has limited tenants hierarchy depth
2. it's hard to implement tenant move (e.g. `259a8820-ef53-40e8-b3a7-a0f78a076c06` is moved under other tenant)

We need to consider other options, e.g. closure table, or just use tenant resolver and fetch all the parents and check settings there.

These ADR pages is done using old-fashioned ADR template. New template is stored here:
https://github.com/hypernetix/hyperspot/blob/main/docs/spec-templates/ADR.md

In the new template we enforce to provide multiple options with their own pros and cons, including potential implementation options and SQLs. Please apply that template here

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
+# PRD
+
+## A. Vision
+
+**Purpose**: Settings Service is a Hyperspot module providing centralized configuration management for multi-tenant SaaS applications. Built on the Hyperspot modular platform using Rust and ModKit, it enables both internal modules and external services to store, retrieve, and manage settings through a type-safe SDK with automatic REST API generation and comprehensive OpenAPI documentation.
+
+The Settings Service acts as a universal configuration management system that supports multiple domain types (Tenant, User) with hierarchical tenant inheritance, role-based access control, and compliance features. It enables services to define setting types with GTS (Global Type System) schemas and manage configuration values across complex organizational hierarchies while maintaining security, auditability, and data integrity through Hyperspot's secure-by-default architecture.
```
Artifizer: I'd also add Resource, where Resource is just an abstract object. We can use it for everything except Tenants and Users.

So then we have (Tenant, User, Resource)

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+
+**Target Users**:
+
+- **Service Developers** - Internal and external developers building services that need configuration management
+- **System Administrators** - Operations teams managing settings across tenant hierarchies
+- **Compliance Officers** - Teams enforcing configuration policies and audit requirements
```
Artifizer: Plus:
- Integrations
- Custom modules 
- Runtime workflows / FaaS functions

that need to store their custom objects associated with tenants, users, resources

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+- Trait-based event generation for audit and notification systems
+- Domain object association (tenant, user)
+- MFA-protected settings for sensitive configurations
+- Generic and explicit value resolution with inheritance chains
+- Soft deletion with configurable retention periods
+- Generic reporting interface for setting values with configurable attribute mappings
+- API for creating new setting types with GTS schema definitions
+- Automatic OpenAPI documentation via ModKit OperationBuilder
+- OData support for pagination, filtering, and field projection
+- Secure-by-default database access with automatic tenant scoping
```
Artifizer:
```suggestion
- Secure-by-default database access with automatic tenant and setting type (GTS) scoping
```

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+
+**ID**: `fdd-settings-service-actor-tenant-management`
+
+<!-- fdd-id-content -->
+**Role**: Provides tenant hierarchy information, user authentication data, and tenant provisioning states. Publishes tenant lifecycle events (create, update, delete) that Settings Service consumes for tenant reconciliation.
+<!-- fdd-id-content -->
+
+#### Event Bus Module
+
+**ID**: `fdd-settings-service-actor-event-bus`
```
Artifizer: This is not an actor as Event Bus Module doesn't communicate with Settings service

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+
+<!-- fdd-id-content -->
+**Role**: Receives audit and notification events from Settings Service when settings or setting values are modified. Distributes events to subscribers for monitoring, alerting, and compliance tracking. Delivers asynchronous events for tenant and user lifecycle changes.
+<!-- fdd-id-content -->
+
+#### GTS Type Registry Module
+
+**ID**: `fdd-settings-service-actor-gts-registry`
+
+<!-- fdd-id-content -->
```
Artifizer: This is not an actor as GTS Type Registry Module doesn't communicate with Settings service

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+
+**ID**: `fdd-settings-service-actor-domain-validation`
+
+<!-- fdd-id-content -->
+**Role**: Optional module that validates existence of domain objects (users) when settings are created or updated. Provides GET endpoints for domain object retrieval and consumes domain object deletion events.
+<!-- fdd-id-content -->
+
+#### Hyperspot ModKit Framework
+
+**ID**: `fdd-settings-service-actor-hyperspot-modkit`
```
Artifizer: This is not an actor, but mandatory compile-time dependency, so I would omit it

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+
+<!-- fdd-id-content -->
+**Role**: Provides the foundational framework for module lifecycle management, REST API generation via OperationBuilder, ClientHub for inter-module communication, secure database access with SecureConn, and RFC-9457 Problem error handling. Enables automatic OpenAPI documentation and type-safe module integration.
+<!-- fdd-id-content -->
+
+#### Hyperspot API Gateway
```
Artifizer: This is not an actor, but mandatory compile-time dependency, so I would omit it

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+- OData support for pagination, filtering, and field projection
+- Secure-by-default database access with automatic tenant scoping
+
+## B. Actors
+
+### Human Actors
+
+#### Service Developer
+
+**ID**: `fdd-settings-service-actor-service-developer`
```
Artifizer: We need implementation status and priorities everywhere:

- [ ] `p1` - `fdd-settings-service-actor-tenant-resolver`

modules/system/settings_service/docs/adrs/0012-fdd-settings-service-adr-inheritance-algorithm.md
```diff
@@ -0,0 +1,49 @@
...
+
+**ADR ID**: `fdd-settings-service-adr-inheritance-algorithm`
+
+## Context and Problem Statement
+
+The Settings Service resolves setting values through tenant hierarchy inheritance chains (10+ levels). We need to determine the algorithm for traversing the hierarchy and resolving inherited values efficiently.
+
+## Decision Drivers
+
+* Must traverse tenant hierarchy up to 10+ levels
```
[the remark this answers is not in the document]
hitengajjar: The performance may degrade drastically for dynamic lookups for higher tenancy tree depth. Hence in my view, we should start with reasonable number and 10 is not a small number either.

Please suggest if we should remote this artificial limit for now?

modules/system/settings_service/docs/adrs/0002-fdd-settings-service-adr-database-technology.md
hitengajjar: Will remove in next update

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+
+<!-- fdd-id-content -->
+**Role**: Resolves tenant context from authentication tokens and provides SecurityContext for all module operations. Enables automatic tenant isolation and multi-tenant access control across all Hyperspot modules.
+<!-- fdd-id-content -->
+
+## C. Functional Requirements
+
+### Setting Type Definition with GTS Schema and Traits
+
+**ID**: `fdd-settings-service-fr-setting-type-definition`
```
Artifizer: We need implementation status and priorities everywhere:

```
- [ ] `p1` - `fdd-settings-service-actor-tenant-resolver`
```

modules/system/settings_service/docs/adrs/0004-fdd-settings-service-adr-event-bus-integration.md
```diff
@@ -0,0 +1,50 @@
+# ADR-0004: Event Bus Integration Pattern
```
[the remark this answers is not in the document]
hitengajjar: Below I have mentioned that we are choosing hyperspot's native event management. If we remove this ADR, then pls suggest how to notify AI to use hyperspot's native event management system. Do we have any system prompt for it?

https://github.com/cyberfabric/cyberfabric-core/pull/477/changes/BASE..c7316f94384fc7b09e635423ec3f2f9b2157fa4d#diff-e795333b77e26d4a800a263a29871c65b703491cc8e1ebdcf77494606df9d607R30

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+
+**Actors**: `fdd-settings-service-actor-service-developer`, `fdd-settings-service-actor-gts-registry`
+<!-- fdd-id-content -->
+
+#### Domain Object Association
+
+**ID**: `fdd-settings-service-fr-domain-objects`
+
+<!-- fdd-id-content -->
+The system MUST associate setting values with domain objects (tenant, user) and support generic values that apply to all domain objects when no explicit value exists. Domain object validation is optional and configurable per setting type through traits.
```
Artifizer: Some ideas regarding the design (not PRD)

I think we can make some simplification:

We can store all the domain objects per tenant and validate the values using tenant key always. In this case, if a `user` or`resource` with given UUID doesn't belong to a tenant there will be nothing bad because:

a. resource/user settings are not hierarchical by design
b. these settings would be per tenant strictly. If the tenants owns it - fine, it can see see the values. Other tenants won't see the values because tenant access check will not pass.

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+
+**Actors**: `fdd-settings-service-actor-service-developer`, `fdd-settings-service-actor-hyperspot-modkit`
+<!-- fdd-id-content -->
+
+#### Security Context Integration
+
+**ID**: `fdd-settings-service-fr-security-context`
+
+<!-- fdd-id-content -->
+The system MUST require security context for all API operations and SDK methods to enable tenant isolation, access control, and audit logging. All database operations MUST use secure database connections with automatic tenant scoping to prevent cross-tenant data access. The system MUST integrate with the platform's tenant resolver for authentication and authorization.
```
Artifizer: The system must implement GTS-scoped access control, so specific setting must be accessible only if given actor has given GTS type allowed in scope (either explicit or using a wildcard)

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+
+**Actors**: `fdd-settings-service-actor-service-developer`, `fdd-settings-service-actor-hyperspot-modkit`
+<!-- fdd-id-content -->
+
+#### Dynamic Domain Type Registration
+
+**ID**: `fdd-settings-service-fr-dynamic-domain-types`
+
+<!-- fdd-id-content -->
+The system MUST support dynamic domain types as GTS entities, allowing new domain types to be registered at runtime without code changes. Each domain type registration MUST provide: (1) domain_id data type specification (e.g., string, uuid), (2) REST API endpoint for validating domain object existence, (3) event type for domain object deletion that Settings Service monitors to automatically remove settings scoped to deleted domain objects. The system MUST support the predefined domain types: TENANT, USER, and allow registration of additional custom domain types through the domain type registry.
```
Artifizer: As of p3 - please add custom Resource as well, not just User.

We probably can have resources associated with Tenant only or with Tenant and User

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+## UC-001: Create Setting Type with GTS Schema
+
+**ID**: `fdd-settings-service-usecase-create-setting-type`
+
+<!-- fdd-id-content -->
+**Actor**: `fdd-settings-service-actor-service-developer`
+
+**Preconditions**:
+
+- User has appropriate permissions
```
Artifizer:
```suggestion
- Actor has appropriate permissions
```

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+**Preconditions**:
+
+- Setting type exists
+- User has read access to tenant
+- Setting type allows generic or explicit queries
+
+**Flow (when inheritance is enabled)**:
+
+1. Service requests setting values for tenant and domain object
+2. System validates tenant read access
```
Artifizer: + GTS permissions

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+
+**Actors**: `fdd-settings-service-actor-tenant-management`, `fdd-settings-service-actor-event-bus`
+<!-- fdd-id-content -->
+
+#### Setting Type API
+
+**ID**: `fdd-settings-service-fr-setting-type-api`
+
+<!-- fdd-id-content -->
+The system MUST provide API endpoints to create new setting types with GTS schema definitions and configurable traits. The API allows developers to define setting types programmatically including domain type, event configuration, setting options, and operation rules.
```
Artifizer: Need to check GTS token claims as well

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+<!-- fdd-id-content -->
+The system MUST support horizontal scaling with stateless API servers. The system MUST handle 100+ concurrent setting types and 10,000+ setting values per type. The system MUST support tenant hierarchies with 100,000+ tenants. Database connection pooling MUST be configured to handle concurrent requests efficiently.
+<!-- fdd-id-content -->
+
+## Data Encryption
+
+**ID**: `fdd-settings-service-nfr-data-encryption`
+
+<!-- fdd-id-content -->
+The system MUST encrypt sensitive setting values at rest using industry-standard encryption algorithms (AES-256 or equivalent). Encryption keys MUST be managed through a secure key management system with key rotation capabilities. Setting types MUST support marking fields as sensitive to enable automatic encryption. Encrypted data MUST be decrypted only when accessed by authorized users with valid security context.
```
Artifizer: CRITICAL: Here is tricky part and I would try to avoid it as much as we can and use the cred store for this kind of artifacts. Settings service can store credential ID, but not any credential, even if encrypted

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+<!-- fdd-id-content -->
+The system MUST comply with data privacy regulations including GDPR, CCPA, and similar frameworks for user-scoped and tenant-scoped settings. The system MUST support data subject access requests (DSAR) to export all settings associated with a user or tenant. The system MUST implement the right to be forgotten by permanently deleting all user data upon request. Personal data MUST be processed with explicit consent and purpose limitation. Data residency requirements MUST be supported for region-specific data storage.
+<!-- fdd-id-content -->
+
+## Network Security
+
+**ID**: `fdd-settings-service-nfr-network-security`
+
+<!-- fdd-id-content -->
+The system MUST use TLS 1.2 or higher for all network communications including API endpoints, inter-module communication, and external service integrations. The system MUST enforce network isolation between tenants to prevent cross-tenant network access. API endpoints MUST be protected by rate limiting and DDoS protection mechanisms. The system MUST support IP allowlisting and network access control lists for restricted environments.
```
Artifizer: I believe the Setting Module itself should not care about TLS as is this API gateway and it's middleware responsibility. It's likely the same for DDoS protection and IP allow list

modules/system/settings_service/docs/PRD.md
```diff
@@ -0,0 +1,875 @@
...
+<!-- fdd-id-content -->
+
+## Default Value Resolution Strategy
+
+**ID**: `fdd-settings-service-prd-context-default-value-resolution`
+
+<!-- fdd-id-content -->
+The service is designed with a "always return a value" philosophy to ensure applications never encounter null or missing configuration errors. The value resolution follows a cascading fallback strategy: (1) explicit value for tenant/domain object, (2) explicit value for tenant/generic object, (3) inherited value from parent tenant hierarchy (if inheritance enabled), (4) default value from setting type's GTS schema. Every setting type MUST define a default value in its GTS schema, which serves as the ultimate fallback. This design ensures predictable behavior, eliminates null-checking in consuming applications, and provides sensible defaults for all configurations. The response metadata indicates the value source (explicit, inherited, generic, or default) for transparency and debugging.
```
Artifizer: Isn't a duplication?

Artifizer (changes requested):

modules/system/settings_service/docs/DESIGN.md
```diff
@@ -0,0 +1,2015 @@
...
+│  │  - TraitsManager (compliance, MFA, inheritance)       │ │
+│  │  - Domain models (SettingType, SettingValue)          │ │
+│  │  - Repository traits (SettingTypeRepo, ValueRepo)     │ │
+│  │  - Event publisher port                               │ │
+│  └────────────────────────────────────────────────────────┘ │
+│                              ↓                               │
+│  ┌────────────────────────────────────────────────────────┐ │
+│  │  Infrastructure Layer (infra/)                         │ │
+│  │  - SeaORM entities and repositories                   │ │
+│  │  - SecureConn for tenant-scoped queries               │ │
```
Artifizer: GTS-scoped queries

modules/system/settings_service/docs/DESIGN.md
```diff
@@ -0,0 +1,2015 @@
...
+<!-- fdd-id-content -->
+
+#### Data Encryption
+
+**ID**: `fdd-settings-service-constraint-data-encryption`
+
+<!-- fdd-id-content -->
+**ADRs**: References PRD NFR-Data Encryption (`fdd-settings-service-nfr-data-encryption`)
+
+MUST encrypt sensitive setting values at rest using AES-256 or equivalent industry-standard encryption algorithms. Encryption keys MUST be managed through a secure key management system with key rotation capabilities. Setting types MUST support marking fields as sensitive to enable automatic encryption. Encrypted data MUST be decrypted only when accessed by authorized users with valid security context.
```
Artifizer: See notes in requirements. I think we do not need to accept secrets in the settings service and use cred store for this

modules/system/settings_service/docs/DESIGN.md
```diff
@@ -0,0 +1,2015 @@
...
+│         │                   ┌──────────────────┐               │
+│         │                   │ ValueResolver    │               │
+│         │                   │ (Inheritance)    │               │
+│         │                   └──────────────────┘               │
+│         │                            │                          │
+│         ↓                            ↓                          │
+│  ┌────────────────┐         ┌──────────────────┐               │
+│  │ Error Mapper   │         │ TenantAccess     │               │
+│  │ (Problem)      │         │ Manager          │               │
+│  └────────────────┘         └──────────────────┘               │
```
Artifizer: plus GTS access manager

modules/system/settings_service/docs/DESIGN.md
```diff
@@ -0,0 +1,2015 @@
...
+**GET /api/settings/v1/types/{type_id}** - Get Setting Type
+
+```json
+// Response: 200 OK
+{
+  "id": "550e8400-e29b-41d4-a716-446655440000",
+  "name": "data.retention",
+  "domain_type": "TENANT",
+  "gts_schema": {
+    "ref": "gts://registry.hyperspot.dev/schemas/data-retention/v1.0.0",
```
Artifizer: Wrong `gts://` reference

modules/system/settings_service/docs/DESIGN.md
```diff
@@ -0,0 +1,2015 @@
...
+   - Setting types reference specific schema version
+   - New versions deployed without breaking existing settings
+   - Validation uses version specified in setting type
+
+4. **Default Values**:
+   - Extract default values from GTS schema definitions
+   - Use as fallback when no explicit/inherited value exists
+   - Ensure all setting types have valid defaults
+
+**Schema Reference Format**: `gts://registry.hyperspot.dev/schemas/data-retention/v1.2.0`
```
Artifizer: Wrong `gts://` reference

Artifizer (changes requested):

Artifizer:
@hitengajjar , my review summary:

```
1. Scope Boundaries (What Doesn't Belong Here)
- Remove 5 ADRs that are global platform decisions, not module-specific
- Remove ~15 NFRs that are platform-level, not module requirements
- Security concerns (TLS, DDoS, encryption) belong at API gateway/platform level
- Don't store credentials - that's credential store's responsibility

2. GTS Type System Issues
- Fix GTS ID format (wrong vendor, wrong pattern)
- Implement GTS-scoped access control throughout
- CRITICAL: Validate GTS access claims to prevent impersonation
- Use GTS schemas for trait definitions
- Base type naming is incorrect

3. Architecture & Dependencies
- Reclassify actors vs dependencies (several modules aren't real actors)
- Add missing actor types (integrations, custom modules, workflows)
- Add GTS access manager component
- Missing fields in domain model (is_locked)

4. Domain Model Simplification
- Add Resource as third domain type alongside Tenant/User
- Simplify domain object handling (always use UUID)
- Simplify validation approach (per-tenant scoping)

5. Tenant Hierarchy Concerns
- Verify approach with @aviator5 (appears in 2 ADRs)
- Materialized path has fundamental issues (depth limits, tenant moves)
- Consider alternative approaches (closure table)
- Remove artificial hierarchy depth limit

6. Documentation Quality
- Update to CyberFabric branding
- Remove obsolete markers
- Add implementation priorities
- Use new ADR template with proper options/tradeoffs

7. Missing Clarifications
- What is "SUBROOT"?
- Need two strategies for default value resolution
- Verify if certain areas need separate GTS access checks
```

modules/system/settings_service/docs/PRD.md
hitengajjar: I think we need to discuss on how to support validation of resource object existence and auto-removal of settings when such resource objects are removed. We need to then introduce a concept of registering the resource type's APIs and event types that settings service module should work with.

In my view, this first version we can keep it simple to support only Tenants and Users and post discussion we can decide on how to support different resource types.

PR #482, opened 2026-02-06
[WiP] service job design: PRD, DESIGN

bsacrobatix:
Summary

  - Add PRD and Technical Design documents for the new Service Jobs module — a lightweight, embedded async job execution
   system that runs within the service process as Tokio tasks
  - The module supports two types of work: restartable jobs (serializable, persisted to DB, survive restarts) and
  non-restartable jobs (in-memory only, lost on restart)
  - Both types share unified timeout, cancellation, progress reporting, and observability primitives

  Key design decisions

  - No external queue infrastructure — uses in-memory channels (P1) and database-backed queuing (P2) instead of
  RabbitMQ/SQS
  - Two-phase implementation: P1 delivers core in-memory execution with retry, DLQ, cancellation, idempotency, and REST
  status API; P2 adds persistence, restart recovery, checkpointing, and GTS-based handler discovery
  - Tenant-scoped isolation enforced on all DB queries and API paths via Secure ORM
  - REST API is read-only (status/result/list queries); job submission, cancellation, and DLQ management remain internal
   Rust APIs only
  - Callbacks defined as handler trait methods (not per-job) so they survive restarts

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive technical design, product requirements, and decision records for the Service Jobs module covering architecture, execution models (in-process and persistent), APIs and lifecycle semantics, restart/recovery, checkpointing, progress streaming, DLQ/retention, tenant isolation, security, and observability.
  * Added a PR review checklist capturing design clarifications, heuristics, and acceptance criteria.


modules/service_jobs/docs/DESIGN.md
```diff
@@ -0,0 +1,1010 @@
...
+All async work uses this module for consistent timeouts, cancellation, and observability. The key difference is persistence:
+
+| Type | Persisted to DB | On Crash/Restart |
+|------|-----------------|------------------|
+| **Restartable** (serializable inputs) | Yes | Job resumes automatically |
+| **Non-restartable** (non-serializable inputs) | No (in-memory) | Job is lost |
+
+Non-restartable jobs skip database writes entirely, avoiding persistence overhead for work that cannot be restarted anyway.
+
+#### Local Worker Execution
```
MikeFalcon77: Running workers as Tokio tasks inside the service is simple, but at scale this design lacks separation of execution and persistence concerns. Standard durable job queues separate execution from API processes so crashes or spikes don’t starve job workers. The architecture ties API, business logic, and job execution tightly together which reduces isolation and complicates backpressure control.

modules/service_jobs/docs/DESIGN.md
```diff
@@ -0,0 +1,1010 @@
...
+On startup, workers detect and recover orphaned restartable jobs:
+
+**Note**: SQL shown for clarity only. Module code must use Secure ORM (`SecureConn` / `AccessScope`). Raw SQL is allowed only in migrations.
+
+```sql
+-- Recover orphaned restartable jobs: reset to pending
+UPDATE service_jobs
+SET status = 'pending', worker_id = NULL, claimed_at = NULL
+WHERE status = 'running'
+  AND claimed_at < NOW() - INTERVAL '5 minutes';
```
MikeFalcon77: Detecting “orphaned” running jobs by claimed_at < now() - 5 minutes is brittle because a legitimately long job will bounce back into pending prematurely. Look like we need to use either heartbeat/lease extensions with visibility timeouts or a lease token that workers renew periodically. The current 5 minute heuristic works as a fallback but without heartbeat it effectively guarantees at least-twice delivery for longer jobs

modules/service_jobs/docs/DESIGN.md
```diff
@@ -0,0 +1,1010 @@
...
+| id | UUID | PK | Job identifier |
+| tenant_id | UUID | NOT NULL | Tenant context |
+| handler_id | VARCHAR | NOT NULL | Handler identifier |
+| status | ENUM | NOT NULL | pending/running/succeeded/failed/canceled/dead_lettered |
+| input | JSONB | NOT NULL | Serialized input |
+| result | JSONB | | Serialized result |
+| checkpoint | JSONB | | Checkpoint data for resuming |
+| retry_policy | JSONB | NOT NULL | Retry configuration |
+| attempt | INT | NOT NULL DEFAULT 0 | Current attempt |
+| idempotency_key | VARCHAR | | Deduplication key; unique constraint on `(idempotency_key, handler_id, tenant_id)` |
```
MikeFalcon77: We declare (idempotency_key, handler_id, etc) uniqueness but do not clearly define how an incoming duplicate submission is handled (409 vs returning existing job vs merging metadata). We can make it robust: returns the existing job result or a stable job ID so clients can safely retry submission without side effects. Please clarify this behavior in your API spec.

MikeFalcon77 (changes requested):

MikeFalcon77 (approved):

PR #527, opened 2026-02-09
OAGW design changes

striped-zebra-dev:
## Summary by CodeRabbit

* **New Features**
  * Added custom header routing via X-OAGW-Target-Host for explicit endpoint selection in multi-endpoint upstreams
  * Enabled disable/re-enable functionality for upstreams and routes
  * Introduced plugin system supporting authentication, guarding, and request/response transformation

* **Documentation**
  * Expanded architecture and deployment guidance documentation
  * Added comprehensive request flow and management operation scenarios
  * Enhanced error handling documentation with RFC 9457 Problem Details format


MikeFalcon77:
I've got security report for this design and dound it interesting:

---

### RISK-01: SSRF — No validation of resolved IP (P0)

* **Risk**: OAGW resolves a hostname from the upstream config and sends a request. If the hostname resolves to a private IP (127.0.0.1, 169.254.169.254, 10.x.x.x, fc00::/7, ::1), OAGW becomes an SSRF proxy to internal services.
* **Impact**: Critical — access to cloud metadata, internal APIs, credential theft.
* **Likelihood**: High — a standard attack against egress proxies.
* **Where**: DESIGN.md §Security Considerations mentions “IP pinning rules” but marks them as “out of scope for this document”. Implementation is deferred to F-P1-008 (Phase 1).
* **Fix**:

  * **Move to P0**: DNS resolve → IP check MUST be in the MVP. This is not hardening, this is baseline security.
  * Implement `validate_resolved_ip(ip)` → block RFC 1918, RFC 6598, RFC 3927, loopback, link-local, multicast, IPv6 ULA and link-local.
  * Validate **after** DNS resolve, before connect. Re-validate on every new connection (DNS rebinding).
  * Optional: resolve DNS in OAGW and pass the IP into the HTTP client via a `resolve` API (reqwest supports this).

### RISK-02: DNS Rebinding (P0)

* **Risk**: An attacker configures upstream `evil.com` → first resolve = `1.2.3.4` (public). OAGW caches it as “allowed”. DNS then changes to `169.254.169.254`. Subsequent requests go to cloud metadata.
* **Impact**: Critical — SSRF to internal services.
* **Likelihood**: Medium — requires DNS control.
* **Where**: DESIGN.md does not mention DNS rebinding at all. The HTTP/2 capability cache (§HTTP Version Negotiation) caches by host/IP, but re-resolve behavior is not specified.
* **Fix**:

  * Validate IP on **every** connection, not only on the first one.
  * Do not cache “allowed” status based on hostname, only on IP.
  * Use a connection pool keyed by IP, not by hostname.

### RISK-03: Missing redirect policy (P0)

* **Risk**: The design **nowhere** describes how HTTP redirects (301, 302, 307, 308) are handled. If reqwest follows redirects by default, then:

  1. Redirect to an internal IP → SSRF
  2. Redirect → Auth headers (Authorization: Bearer) leak to another domain
  3. Redirect loop → DoS
* **Impact**: Critical (SSRF + credential leak).
* **Likelihood**: High — redirects are common.
* **Where**: **Missing** from the design. No section, no ADR, not mentioned in Security Considerations.
* **Fix**:

  * **Disable redirect following** by default. Return the redirect response to the client.
  * If following redirects is required, apply the same SSRF checks to the redirect target.
  * **Never** forward Authorization headers when redirecting to a different domain.
  * Limit max redirects (3–5).
  * Add a dedicated section to DESIGN.

### RISK-04: Header Smuggling via X-OAGW-Target-Host (P0)

* **Risk**: A new PR introduces `X-OAGW-Target-Host` to select an endpoint. The caller supplies the value, and OAGW compares it case-insensitively with endpoint hosts. However:

  1. Unicode, IDN, or punycode may bypass string comparison
  2. If OAGW uses this value for the downstream `Host` header, host header injection is possible
  3. The design says “stripped before forwarding”, but this is an **invariant**, not an enforced point
* **Impact**: High — routing to unintended endpoints, potential SSRF.
* **Likelihood**: Medium.
* **Where**: DESIGN.md §Alias Resolution, §Headers Transformation, PR diff with new scenarios.
* **Fix**:
* **Note**: I see basic things but we need to implement ASCII-only, punycode/IDN normalizaiton for this header

  * Validate `X-OAGW-Target-Host`: ASCII-only, no `@`, no `/`, no `:`, no null bytes, no newlines.
  * Compare normalized values (lowercase, punycode-decoded).
  * **Explicitly document** that `Host`/`:authority` is ALWAYS set to `upstream.endpoint.host`, never from `X-OAGW-Target-Host`.
  * Add negative tests: X-OAGW-Target-Host with unicode, punycode, port, userinfo, path.

### RISK-05: Tenant Isolation — Cache Poisoning (P1)

* **Risk**: Cache keys in ADR Data Plane Caching:

  * `upstream:{tenant_id}:{alias}` — OK, includes tenant
  * `route:{upstream_id}:{method}:{path}` — **does NOT include tenant_id**
  * `plugin:{plugin_id}` — **does NOT include tenant_id**
  * CP L1 cache does not describe keys at all
    If `upstream_id` accidentally collides or an upstream is shared between tenants, the route cache may return another tenant’s route.
* **Impact**: High — cross-tenant data leak, routing to wrong config.
* **Likelihood**: Medium — depends on shared upstream implementation.
* **Where**: adr-data-plane-caching.md §Cache Keys.
* **Fix**:

  * **All** cache keys MUST include `tenant_id` (or scope chain).
  * `route:{tenant_id}:{upstream_id}:{method}:{path}`
  * `plugin:{tenant_id}:{plugin_id}`
  * CP L1 cache must be keyed by `(tenant_id, upstream_id)`.

### RISK-06: Confused Deputy — cred_store Access (P1)

* **Risk**: Auth config contains `secret_ref: "cred://partner-openai-key"`. OAGW calls cred_store on behalf of whom? The design says “cred_store checks if the secret is accessible to the current tenant”, but:

  1. Who authenticates to cred_store? If OAGW uses a service account, cred_store may not enforce tenant scope.
  2. If tenant A sets `secret_ref: "cred://tenant-b-secret"` and cred_store does not check properly, secrets leak.
* **Impact**: Critical — cross-tenant secret theft.
* **Likelihood**: Medium — depends on cred_store implementation.
* **Where**: DESIGN.md §Secret Access Control — describes desired behavior, but not the enforcement mechanism.
* **Notes**: Mostly covered

  * OAGW MUST pass `tenant_id` and `principal_id` to cred_store on every request.
  * cred_store MUST enforce access control, but OAGW MUST add defense in depth: validate that `secret_ref` belongs to the tenant BEFORE calling cred_store.
  * Add audit logging for every cred_store access.
  * Add an integration test: tenant A cannot read tenant B’s secret.

### RISK-07: Secret leakage in error responses (P1)

* **Risk**: If cred_store returns an error containing secret material, or if an upstream returns a response that includes sent credentials in the body, OAGW may pass this to the client.
* **Impact**: High — credential exposure.
* **Likelihood**: Medium.
* **Where**: DESIGN.md §Error Handling — passthrough of upstream errors “status + headers + body without modification”. §Audit Logging — “No secrets in logs”. No mechanism for response body scanning.
* **Fix**:

  * cred_store errors MUST be replaced with a generic `500 SecretNotFound`, without details.
  * Upstream error passthrough is OK, but redact any OAGW-injected headers (Authorization, API keys) from error details.
  * On auth plugin failure, do not include secret material in Problem details.

### RISK-08: Request Body DoS — Decompression Bombs (P1)

* **Risk**: There is a 100MB hard body limit. But with `Transfer-Encoding: chunked` and `Content-Encoding: gzip`, a compressed 1KB payload can decompress into gigabytes.
* **Impact**: High — OOM, crash.
* **Likelihood**: Medium.
* **Where**: DESIGN.md §Body Validation Rules — “Hard limit 100MB; reject before buffering”. No mention of Content-Encoding.
* **Fix**:

  * Enforce limits on **decompressed** size, not just wire size.
  * Or do not decompress at all (proxy as-is) and enforce limits on wire size.
  * Reject unsupported Content-Encoding values.
  * Add this to Body Validation Rules.

### RISK-09: Starlark Plugin Escape — Credential Access (P1)

* **Risk**: Starlark plugins have access to `ctx.request.headers` (read/write). After auth plugin injection, headers contain `Authorization: Bearer <secret>`. A transform plugin (on_request phase) runs AFTER auth and can read and exfiltrate it via `ctx.log.info("got token", ctx.request.headers.get("Authorization"))`.
* **Impact**: High — secret leak via plugin logs.
* **Likelihood**: High — per design plugin chain: Auth → Guard → Transform(request).
* **Where**: DESIGN.md §Plugin Execution Order, §Starlark Context API.
* **Fix**:

  * `ctx.request.headers` exposed to Starlark MUST **hide** security-sensitive headers (Authorization, Proxy-Authorization, Cookie).
  * Or inject auth AFTER transform plugins (architectural change).
  * Or add a redaction layer to `ctx.log.*` to strip known secret patterns.
  * Enforce log size limits per plugin call (already mentioned in F-P1-005, but must be enforced).

### RISK-10: IDOR via Management API (P1)

* **Risk**: The Management API accepts GTS IDs as UUIDs. If tenant A guesses the UUID of tenant B’s upstream and Secure ORM does not enforce scope on every query, this is IDOR.
* **Impact**: High — read or modify чужая configuration.
* **Likelihood**: Low to Medium (UUIDs are hard to guess, but UUIDv7 is predictable).
* **Where**: DESIGN.md §Management API, F-P1-003.
* **Fix**:

  * SeaORM Secure ORM (`SecureConn`) with tenant scope on EVERY query, already planned (F-P1-002, F-P1-003).
  * Move enforcement from P1 to P0: deny-by-default scope.
  * Add a test: request for another tenant’s upstream UUID → 404 (not 403, to avoid leaking existence).

### RISK-11: Rate Limit Bypass via Multiple Tenants (P2)

* **Risk**: Default scope = “tenant”. If an attacker creates N child tenants, each gets its own rate limit bucket. Effective limit = N * configured_rate.
* **Impact**: High — billing abuse, upstream overload.
* **Likelihood**: Medium — depends on tenant creation policy.
* **Where**: DESIGN.md §Rate Limiting, ADR Rate Limiting.
* **Fix**:

  * Enforce aggregate rate limits per upstream (across all tenants), a “global” scope.
  * Ancestor `enforce` rate limits must be per upstream instance, not per tenant.
  * Clarify scope hierarchy: `min(global, ancestor.enforce, tenant)`.

### RISK-12: SQL Injection via JSONB Queries (P1)

* **Risk**: Common Queries use JSONB path expressions with string interpolation:

  ```sql
  AND $3 LIKE (match ->'http'->>'path' || '%')
  ```

  If `path` in JSONB contains SQL wildcards (`%`, `_`), pattern matching can be corrupted. Not classic SQL injection, but logic bypass.
* **Impact**: Medium — route matching corruption.
* **Likelihood**: Low.
* **Where**: DESIGN.md §Common Queries — “SQL in DESIGN is illustrative”, but it defines semantics.
* **Fix**:

  * Escape `%` and `_` in values coming from JSONB before LIKE.
  * Or use prefix comparison (`starts_with`) instead of LIKE.
  * F-P1-003 already addresses “no raw SQL”, ensure this also covers LIKE injection.

### RISK-13: Cache Invalidation Race — Stale Config Serving (P2)

* **Risk**: Config cache (L1/L2) has no TTL in L1. If an admin disables an upstream (enabled=false) for emergency shutdown, L1 cache may continue serving requests until eviction.
* **Impact**: Medium — bypass of emergency shutoff.
* **Likelihood**: Medium.
* **Where**: adr-data-plane-caching.md — “L1: No TTL (evicted by LRU)”.
* **Fix**:

  * L1 cache MUST have a bounded TTL (30–60s) in addition to LRU.
  * Or implement synchronous invalidation from DP to CP on security-critical changes (enabled or disabled toggle).
  * CP L1 cache notification mechanism is not described (“notified by DP or periodic sync”).

### RISK-14: URL Parsing — `@` in userinfo, scheme confusion (P1)

* **Risk**: Upstream endpoint `host` field allows `format: hostname`. But:

  * `evil@internal.host` if the URL parser treats it as userinfo
  * IDN homograph: `аpi.openai.com` (Cyrillic ‘а’) vs `api.openai.com`
  * Non-standard ports in alias: `api.openai.com:8443`, alias comparison may be inconsistent
* **Impact**: High — SSRF, credential routing to the wrong host.
* **Likelihood**: Low to Medium.
* **Where**: DESIGN.md §Upstream Schema — `host` is `format: hostname` but no explicit normalization.
* **Fix**:

  * Strict ASCII-only validation for `host`. Reject non-ASCII, `@`, `#`, `?`, `/`.
  * Normalize host: lowercase, punycode → reject or convert.
  * Validate at the management API (write time), not at proxy time.

### RISK-15: Proxy Headers Not Stripped (P1)

* **Risk**: A caller can send `X-Forwarded-For`, `X-Real-IP`, `X-Forwarded-Host`. If the upstream trusts these headers, an attacker can spoof the source IP.
* **Impact**: Medium — bypass of IP-based access control at the upstream.
* **Likelihood**: Medium.
* **Where**: DESIGN.md §Headers Transformation — lists hop-by-hop headers to strip but does NOT include `X-Forwarded-*`, `X-Real-IP`, `Forwarded`.
* **Fix**:

  * Add `X-Forwarded-For`, `X-Forwarded-Host`, `X-Forwarded-Proto`, `X-Real-IP`, `Forwarded` to the strip list.
  * Or only pass through explicitly allowlisted headers (default `passthrough: none` is good, but verify enforcement).

MikeFalcon77:
About RISK report I have a new data. It's comparision with Kong and Tyk. Looks like we are ok in most ticks:

# OAGW Security Risks: Comparison with Kong and Tyk

## RISK-01: SSRF via IP (P0) — Upstream URL resolves to private IP (10.x, 169.254.x, etc.)

| | Kong | Tyk | Verdict for OAGW |
|---|---|---|---|
| **Protection** | **None built-in.** `ip-restriction` plugin filters **client** IPs, not upstream targets. Admin enters upstream hostname/IP via Admin API. No validation that target doesn't resolve to private IP. | **None built-in.** Upstream target_url is set by administrator. JSON schema validation is for request payload, not upstream IP. | **Real risk, but industry norm is not to protect.** Kong and Tyk trust the administrator. OAGW can add optional `deny_private_ip` flag as defence-in-depth, but this is **not P0 but P2** — downgrade. |

## RISK-02: DNS Rebinding (P2, downgraded) — Upstream hostname initially resolves safely, later resolves to internal IP

| | Kong | Tyk | Verdict for OAGW |
|---|---|---|---|
| **Protection** | **Not implemented.** No DNS pinning or re-validation. Nginx resolves DNS according to TTL. | **Not implemented.** Go net/http standard DNS resolution. There was a PR on DNS caching (#2040) — but for performance, not security. | **Confirmed: P2.** No gateway implements this. OAGW is egress-only, upstream targets are admin-configured. Keep P2. |

## RISK-03: Redirect Following (P1) — CP's HTTP client follows 3xx redirect to internal service

| | Kong | Tyk | Verdict for OAGW |
|---|---|---|---|
| **Protection** | **Does not follow redirects.** Kong is a transparent proxy (nginx-based). 3xx from upstream is transparently returned to client. Kong does not follow redirects on behalf of the client. | **Does not follow redirects.** Tyk is a transparent proxy. 3xx from upstream is proxied back to client. "Looping" is internal URL rewrite, not following upstream redirects. | **Confirmed: OAGW must NOT follow redirects** (redirect_policy: `none` by default, like Kong/Tyk). If follow is enabled — validate destination IP. Matches industry. |

## RISK-04: Host Header / Hop-by-Hop (P1) — Client-supplied headers (Host, Connection, Transfer-Encoding) forwarded to upstream

| | Kong | Tyk | Verdict for OAGW |
|---|---|---|---|
| **Protection** | **Yes.** `preserve_host=false` (default) — Kong sets Host = upstream hostname. `preserve_host=true` — forwards original Host. `host_header` on upstream entity — override. Hop-by-hop: nginx removes Connection, Transfer-Encoding when proxying. [Issue #2202](https://github.com/Kong/kong/issues/2202) — hop-by-hop handling improvement discussed. | **Partial.** Tyk sets Host to target_url hostname by default. preserve_host_header in API definition. Go net/http handles hop-by-hop headers by default. | **Standard practice, OAGW must implement.** Strip hop-by-hop headers (Connection, Transfer-Encoding, Upgrade). Set Host = upstream hostname. Matches Kong/Tyk. |

## RISK-05: Cache Poisoning via Missing Tenant (P0) — Cache keys like `route:{upstream_id}:{method}:{path}` without tenant_id

| | Kong | Tyk | Verdict for OAGW |
|---|---|---|---|
| **Protection** | **Implicit isolation.** Kong Enterprise Workspaces — each workspace has its own entities (Services, Routes). Entity IDs are globally unique (UUID). Cache by entity ID is automatically isolated. `db_cache_ttl`, `db_cache_neg_ttl`, `db_cache_stale_ttl` — configurable. | **Implicit isolation.** Tyk — APIs bound to org_id. Redis cache per-API (API ID = key). API IDs are unique within organization. Cache flush — per API ID. | **Confirmed P0 for OAGW.** Kong/Tyk solve this through UUID-based entity isolation. OAGW cache keys use `upstream_id` (UUID) — this already contains implicit tenant isolation. **BUT**: key `route:{upstream_id}:{method}:{path}` — if upstream_id is scoped to tenant, then OK. Need to verify that upstream_id is always scoped to tenant. |

## RISK-06: Credential Store Leak in Logs (P0) — Secrets from cred_store appear in error messages or logs

| | Kong | Tyk | Verdict for OAGW |
|---|---|---|---|
| **Protection** | **Partial.** OSS — credentials in plain text in DB. Enterprise — Vault integration (HashiCorp, AWS SM, GCP SM, env vars). Secrets referenced via `{vault://...}` URI — config stores reference, not value. But: key-auth plugin doesn't encrypt even in Enterprise. Admin API can return credentials in plain text. | **Yes.** `enable_key_logging: false` (default) — masks API keys in logs. Secrets masked as `****`. Tyk Dashboard RBAC restricts access to keys. | **Confirmed P0.** Both platforms had issues. Tyk is better (default masking). OAGW must: 1) never log secrets, 2) vault references instead of values, 3) redact in error responses. |

## RISK-07: Secret Leak in Error Responses (P1) — Upstream URL or auth tokens appear in error messages returned to caller

| | Kong | Tyk | Verdict for OAGW |
|---|---|---|---|
| **Protection** | **Partial.** Kong returns standard errors (`502 Bad Gateway`, `503 Service Unavailable`). Upstream URL **not** included in response body by default. But error logs may contain upstream details. Admin API can expose configuration. | **Partial.** Tyk returns standard error codes. `MaxResponseBodySize` = 0 (no limit). Error body from upstream is proxied as-is. Custom error templates supported. | **Standard practice.** OAGW must: generic error messages for client, upstream details only in internal logs. Kong/Tyk both follow this pattern (with caveats). |

## RISK-08: Decompression Bomb / Body Limit (P1) — No limit on upstream response body causes OOM

| | Kong | Tyk | Verdict for OAGW |
|---|---|---|---|
| **Protection** | **Yes (request).** `Request Size Limiting` plugin. `nginx_http_client_max_body_size` (default: 0 = disabled). **Response:** `Response Size Limiting` plugin (community, by Content-Length header). Nginx buffer settings. Decompression bomb — no special protection. | **Yes (both directions).** `MaxRequestBodySize` — 413 on exceed. `MaxResponseBodySize` — 500 on exceed (only works with Response Body Transform middleware). Decompression bomb — no special protection. | **Standard practice.** Both platforms limit by Content-Length, **but do NOT protect against decompression bomb** (gzip bomb). OAGW must: max body size on request and response, + streaming limit on decompressed size during decompression. |

## RISK-09: Starlark Plugin Credential Access (P1) — Custom plugins can extract and exfiltrate secrets from the ctx object

| | Kong | Tyk | Verdict for OAGW |
|---|---|---|---|
| **Protection** | **No sandbox.** Lua plugins have full access to Kong PDK (kong.ctx, kong.service, kong.request). Lua sandbox exists (`untrusted_lua`), but by default plugins run with full access. Enterprise: can restrict which plugins are allowed. OSS: credentials store (key-auth) accessible from plugin code. | **No sandbox.** Go plugins — native code, full access. JS plugins (JSVM) — receive `request` and `session` objects. gRPC plugins — full API via Tyk protocol. No isolation of secrets from plugin code. | **OAGW is in a better position.** Starlark sandbox is already more secure than Lua/Go/JS in competitors. But ctx API must provide **only** needed data. Don't pass raw credentials to ctx. Only `Authorization` header injection after plugin chain. |

## RISK-10: IDOR — Tenant A accesses Tenant B's upstream/route/plugin by guessing UUID

| | Kong | Tyk | Verdict for OAGW |
|---|---|---|---|
| **Protection** | **Enterprise: Workspaces + RBAC.** Each workspace — isolated entities. Admin API scoped to workspace. RBAC roles: Super Admin, Admin, Read-Only. Cross-workspace access requires explicit permission. OSS: no multi-tenancy. | **Dashboard: Organizations + RBAC.** Each org — isolated unit. API keys bound to org_id. Dashboard RBAC: Read/Write/Deny per section. org_id in API definition ensures isolation. | **Standard practice.** Both platforms solve through workspace/org scoping + RBAC. OAGW: every CRUD request MUST check `tenant_id` ownership. `WHERE tenant_id = ?` in every SQL query. UUID is not enough — tenant filter required. |

## RISK-11: Rate Limit Bypass via Multiple Tenants (P1) — Attacker creates N tenants to get N x rate_limit

| | Kong | Tyk | Verdict for OAGW |
|---|---|---|---|
| **Protection** | **No aggregate limit.** Rate Limiting plugin — per consumer/route/service. No cross-workspace aggregate. Kong Plus/Enterprise: no global aggregate across tenants. | **Yes, org-level.** Tyk supports Organization rate limits (aggregate across all APIs in org). MDCB distributes org limits to slave gateways. API-level rate limit — aggregate across all consumers. DRL (Distributed Rate Limiter) — across gateway instances. | **Tyk is better.** OAGW must: 1) per-tenant rate limit, 2) **global aggregate** rate limit per upstream (all tenants combined). Tyk's org-level limits are a good pattern. |

## RISK-12: SQL Injection via Config Values (P2) — Plugin config or upstream URL containing SQL injection payload

| | Kong | Tyk | Verdict for OAGW |
|---|---|---|---|
| **Protection** | **Implicit (ORM).** Kong uses PostgreSQL/Cassandra via its own DAO layer. Entity schemas are validated. **Explicit:** Injection Protection plugin (Enterprise 3.9+) — SQL injection detection in request headers/body/path. But this is for client requests, not admin config values. | **Implicit (ORM).** Tyk uses MongoDB/SQL via standard Go database drivers with parameterized queries. Request Validation middleware — JSON schema validation. | **Standard practice, P2 confirmed.** Both platforms rely on ORM/parameterized queries. OAGW uses modkit-db — must use parameterized queries. Config values must pass schema validation. Not critical if ORM is used correctly. |

## RISK-13: L1 Cache Without TTL (P1) — Stale config served indefinitely from LRU cache

| | Kong | Tyk | Verdict for OAGW |
|---|---|---|---|
| **Protection** | **Has TTL.** `db_cache_ttl` (default: 0 = never expire). `db_cache_neg_ttl` — negative cache TTL. `db_cache_stale_ttl` — stale resurrection TTL. Event-based invalidation: Admin API write -> invalidation event -> all nodes flush. Cluster invalidation events via DB polling. | **Has TTL.** `cache_timeout` per API. Redis-based cache with TTL. Cache flush API endpoint per API. Upstream-controlled TTL via `x-tyk-cache-action-set-ttl` header. | **OAGW ADR states "L1: No TTL, evicted by LRU" — this deviates from industry.** Kong also defaults to TTL=0 (no expire), BUT has event-based invalidation. OAGW also has invalidation on write — this compensates for lack of TTL. Recommendation: add optional safety-net TTL (e.g. 1 hour) as fallback. |

## RISK-14: URL/Hostname Parsing Differences (P2) — Different URL parsers between validation and actual HTTP call lead to bypass

| | Kong | Tyk | Verdict for OAGW |
|---|---|---|---|
| **Protection** | **Implicit.** Kong — nginx URL parsing (single parser). Target = `host:port`, validated on creation via Admin API. No dual-parser problem — nginx parses, validates, and proxies. | **Implicit.** Tyk — Go `net/url` parsing (single parser). `target_url` validated on API creation. Go standard library URL parsing. | **P2 confirmed.** Both platforms use a single parser. OAGW: if using `hyper` for HTTP calls — use `hyper::Uri` for both validation and request. One parser = no divergence. |

## RISK-15: X-Forwarded / Proxy Header Injection (P1) — Client injects X-Forwarded-For, X-Real-IP to appear as internal

| | Kong | Tyk | Verdict for OAGW |
|---|---|---|---|
| **Protection** | **Yes.** `trusted_ips` — CIDR blocks of trusted IPs. Trusted IP -> forward X-Forwarded-*. Untrusted -> Kong **overwrites** with own values. `real_ip_header`, `real_ip_recursive`. Can disable injection: header config = `"off"`. [Issue #11752](https://github.com/Kong/kong/issues/11752) — selective forwarding. | **Partial.** Tyk adds `X-Forwarded-For` header. No explicit trusted_ips mechanism like Kong. Client-supplied headers can be overwritten by middleware. | **Standard practice.** Kong is better (trusted_ips). OAGW is an egress gateway, so X-Forwarded is less critical (no external clients directly). But for internal callers: strip untrusted X-Forwarded-* headers. |

---

## Summary Table

| Risk | P | Kong | Tyk | Verdict |
|---|---|---|---|---|
| RISK-01 SSRF/IP | P0->**P2** | None | None | **Downgrade.** Industry doesn't protect. Optional deny_private_ip. |
| RISK-02 DNS Rebind | P2 | None | None | **Confirmed P2.** Nobody does it. |
| RISK-03 Redirects | P1 | Doesn't follow | Doesn't follow | **Confirmed.** OAGW must not follow (default). |
| RISK-04 Host/Hop | P1 | Yes | Partial | **Confirmed.** Standard proxy responsibility. |
| RISK-05 Cache tenant | P0->**P1** | Implicit (UUID) | Implicit (org_id) | **Downgrade if upstream_id is scoped.** Verify. |
| RISK-06 Cred logs | P0 | Partial | Yes (default mask) | **Confirmed P0.** OAGW must mask. |
| RISK-07 Error secrets | P1 | Partial | Partial | **Confirmed.** Generic errors to client. |
| RISK-08 Body bomb | P1 | Yes (partial) | Yes (partial) | **Confirmed.** + decompression limit needed. |
| RISK-09 Plugin creds | P1 | No sandbox | No sandbox | **OAGW is better** (Starlark). But restrict ctx. |
| RISK-10 IDOR | P1 | Enterprise RBAC | Org RBAC | **Confirmed.** tenant_id in every query. |
| RISK-11 Rate bypass | P1 | No aggregate | Yes org-level | **Confirmed.** OAGW needs per-upstream aggregate. |
| RISK-12 SQLi | P2 | ORM | ORM | **Confirmed P2.** Parameterized queries. |
| RISK-13 L1 no TTL | P1 | TTL=0 + events | TTL + flush API | **Confirmed.** Add safety-net TTL. |
| RISK-14 URL parse | P2 | Single parser | Single parser | **Confirmed P2.** Single parser = OK. |
| RISK-15 X-Forwarded | P1 | Yes (trusted_ips) | Partial | **Confirmed.** Strip for egress. |

## Key Takeaways

1. **RISK-01 and RISK-02 — downgrade.** Neither Kong nor Tyk validate upstream IP. This is the administrator's / network layer's responsibility.

2. **RISK-05 — requires verification.** If `upstream_id` (UUID) is already bound to tenant — cache poisoning is impossible. Need to verify in code.

3. **RISK-09 — OAGW is in a better position** than competitors. Starlark sandbox > Lua/Go/JS without sandbox. But need to restrict what ctx exposes.

4. **RISK-11 — Tyk is better** with org-level aggregate rate limits. OAGW needs an analogous mechanism.

5. **RISK-13 — add safety-net TTL.** Kong also defaults to TTL=0, but has event-based invalidation. OAGW also has this — but safety-net TTL protects against missed invalidation events.

6. **Originally there were 3 P0 risks. After comparison, 1 real P0 remains — RISK-06 (credential masking in logs).** The rest are P1/P2.

Artifizer (changes requested):

modules/system/oagw/DESIGN.md
```diff
@@ -93,12 +159,56 @@
...
+**Deployment**: Modkit framework handles both single-executable and microservice deployment modes, wiring components via in-process function calls or RPC transparently.
+
+For detailed architectural decisions, see:
+
+- [ADR: Component Architecture](./docs/adr-component-architecture.md)
+- [ADR: Request Routing](./docs/adr-request-routing.md)
+
+### Terminology Note
+
+**Crate Names vs. Component Terminology**: Due to historical naming decisions, crate names do not align with component terminology:
```
mattgarmon: For module stubs, we have `oagw-cp/src/control_plane.rs` and `oagw-dp/src/data_plane.rs`
In the cargo files, they align with the descriptions described here.

Files should be renamed to be consistent.

modules/system/oagw/oagw-cp/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "cf-oagw-cp"
+version.workspace = true
+edition.workspace = true
+description = "Data Plane implementation - proxy orchestration and plugin execution"
```
Artifizer: This is Control Plane

modules/system/oagw/DESIGN.md
```diff
@@ -1,279 +1,476 @@
...
+├── oagw-api/          # API Handler implementation
+│                      # Implements: request routing, incoming auth/rate limiting
+│                      # Routes: /proxy/* → CP, /upstreams/* → DP
+│
+├── oagw-cp/           # Data Plane implementation
+│                      # Implements: ControlPlaneService trait
+│                      # Responsibilities: proxy orchestration, plugin execution
+│                      # Includes: built-in plugins (auth, guard, transform)
+│
+└── oagw-dp/           # Control Plane implementation
```
Artifizer: CP/DB confusion

PR #536, opened 2026-02-10
docs(usage_collector): PRD

binarycode:
## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive PRD for the Usage Collector covering purpose, goals, actors, scope, functional and non-functional requirements, ingestion paths (SDK, agent, REST), storage/backends, metric semantics, rate limiting, backfill/reconciliation, type system, observability, and operational workflows.
  * Includes concrete use cases, acceptance criteria, dependencies, risks, assumptions, open questions, governance contracts, and audit/operational controls.
  * Documentation-only change; no production code or public API declarations were modified.

### DIFF
--- a/modules/system/usage_collector/docs/PRD.md
+++ b/modules/system/usage_collector/docs/PRD.md
@@ -4,13 +4,13 @@
 
 ### 1.1 Purpose
 
-A centralized platform service for collecting, storing, and distributing usage data from all sources, enabling accurate metering for billing, quota enforcement, monitoring, and analytics with tenant isolation and type-safe usage definitions.
+A centralized usage ledger (metering system) for reliably collecting, persisting, and exposing raw usage records from all platform sources, with exactly-once semantics, tenant isolation, and type-safe usage definitions. The Usage Collector acts as the authoritative record of resource consumption — it does not aggregate, interpret, or act on the data. Downstream systems (billing, quota enforcement, monitoring, analytics) consume raw usage records and apply their own aggregation rules, policies, and business logic.
 
 ### 1.2 Background / Problem Statement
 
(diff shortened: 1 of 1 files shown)


modules/system/usage_collector/docs/PRD.md
```diff
@@ -0,0 +1,798 @@
...
+The system **MUST** support sustained ingestion of at least 600,000 usage records per minute (10,000 events per second) under normal operation.
+
+**Threshold**: >= 600,000 records/minute (10,000 events/sec) sustained
+**Rationale**: High-volume services (LLM Gateway, API Gateway) generate significant event throughput.
+
+#### Ingestion Latency
+
+- [ ] `p1` - **ID**: `fdd-uc-req-ingestion-latency`
+
+The system **MUST** complete usage record ingestion within 200ms (p95) under normal load.
```
Artifizer: Here we can add a requirement that Query operations or retention jobs must not impact Ingestion latency.

In the DESIGN, I'd consider design like:
- usage is always inserted into operational database
- usage history is stored and maintained separately in historical database (with different plugins, "column" store DB, etc)

modules/system/usage_collector/docs/PRD.md
```diff
@@ -0,0 +1,798 @@
...
+**Actor**: `fdd-uc-actor-quota-enforcement`
+
+**Main Flow**:
+1. Quota Enforcement System queries UC API for current usage
+2. UC returns aggregated usage for tenant and usage type
+3. Quota Enforcement System compares against tenant limits
+4. Quota Enforcement System triggers enforcement if threshold exceeded
+
+**Postconditions**:
+- Tenant usage enforced before exceeding quota; no over-consumption
```
nonameffh: In this scenario quota enforcement and usage reporting are separated and client should know what metric is quotable. In Constructor, we assumed that UC is the entry point for all usage operations, and if a metric has some attribute identifying that it's quotable metric, UC will perform a quota change operation (with acquisition if not previously acquired on client side) as a part of usage reporting. 
In other words - usage of every metric should be reported to UC but not all trackable metric are quotable and if they're - it's a part of usage collection flow to decrease the possibility of error on the client side.
From my perspective such flow can be supported too if we extend a bit model of "storage" to a plugin-based usage handler

PR #539, opened 2026-02-10
chore: enable `docs.rs.all-features` for published crates

fluiderson:
According to https://docs.rs/about/metadata, this is disabled by default. Hence, for example, [`modkit::bootstrap` is inaccessible](https://docs.rs/cf-modkit/0.2.1/modkit/bootstrap/index.html).

## Summary by CodeRabbit

* **Chores**
  * Updated documentation build configuration across all packages to enable building with all feature flags on docs.rs, ensuring comprehensive documentation coverage.

### DIFF
--- a/libs/modkit-auth/Cargo.toml
+++ b/libs/modkit-auth/Cargo.toml
@@ -10,6 +10,7 @@ description = "ModKit authentication library"
 readme = "README.md"
 keywords = ["cyberfabric", "cyberfabric-modkit"]
 categories = ["development-tools"]
+metadata.docs.rs.all-features = true
 
 [lib]
 name = "modkit_auth"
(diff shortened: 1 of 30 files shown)


Artifizer:
Hi, thank you for the PR. The problem is clear and right now this PR is a symptom of missing workspace automation.

This is what is recommended to do in this case

1) Create a tool that enforces workspace metadata across crates.

Example idea:
`cargo xtask fix-docsrs`

This:
- scans workspace members
- inserts missing metadata automatically

This is how large Rust mono-repos manage Cargo.toml drift, and our repo is big enough to justify this.

2) The complete idiomatic config is usually:

```
[package.metadata.docs.rs]
all-features = true
rustdoc-args = ["--cfg", "docsrs"]
```

Why this matters:
Many crates use:

#[cfg(docsrs)]

to:
- enable doc-only modules
- disable platform-specific code
- avoid build failures on docs.rs

Right now you only did half of the common pattern.

3) Some crates inside libs/ probably SHOULD NOT enable all features

@MikeFalcon77 , @Bechma , please help here

Bechma:
For now the all features flag can work to document everything. Although we have in modkit-macros certain traits that are private. So please, make sure that those remain undocumented and hidden.

PR #619, opened 2026-02-13
docs(usage-collector): PRD and DESIGN

capybutler:
## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive technical design and product specification documents for the Usage Collector system, covering architecture, APIs, multi-tenant support, rate limiting, and backfill operations.
  * Added technical design and product specification documents for the Usage Query system, detailing read-path architecture, server-side aggregation, and pluggable backend support.

### DIFF
--- a/modules/system/usage-collector/docs/DESIGN.md
+++ b/modules/system/usage-collector/docs/DESIGN.md
@@ -1,0 +1,1010 @@
+# Technical Design — Usage Collector
+
+
+<!-- toc -->
+
+- [1. Architecture Overview](#1-architecture-overview)
+  - [1.1 Architectural Vision](#11-architectural-vision)
+  - [1.2 Architecture Drivers](#12-architecture-drivers)
+  - [1.3 Architecture Layers](#13-architecture-layers)
+- [2. Principles & Constraints](#2-principles--constraints)
+  - [2.1 Design Principles](#21-design-principles)
+  - [2.2 Constraints](#22-constraints)
+- [3. Technical Architecture](#3-technical-architecture)
+  - [3.1 Domain Model](#31-domain-model)
+  - [3.2 Component Model](#32-component-model)
+  - [3.3 API Contracts](#33-api-contracts)
+  - [3.4 Internal Dependencies](#34-internal-dependencies)
+  - [3.5 External Dependencies](#35-external-dependencies)
+  - [3.6 Interactions & Sequences](#36-interactions--sequences)
+  - [3.7 Database schemas & tables](#37-database-schemas--tables)
+  - [3.8 Observability](#38-observability)
+- [4. Additional context](#4-additional-context)
+- [5. Traceability](#5-traceability)
+
+<!-- /toc -->
+
+## 1. Architecture Overview
+
+### 1.1 Architectural Vision
+
+Usage Collector follows the ModKit Gateway + Plugins pattern. The gateway module (`usage-collector`) is the single centralized service that receives usage records, enforces tenant isolation, and delegates all storage operations to the active plugin. Backend-specific persistence and query logic for ClickHouse or TimescaleDB is encapsulated in storage plugins that register via the GTS type system and are selected by operator configuration. The gateway contains no backend-specific logic.
+
+The SDK crate (`usage-collector-sdk`) defines two trait boundaries: `UsageCollectorClientV1` for usage sources and consumers, and `UsageCollectorPluginClientV1` for storage backend implementations. When a usage source emits a record, the SDK persists it to the source's local database within the same transaction as the caller's domain operation (transactional outbox pattern). The SDK initializes and owns an `OutboxHandle` whose background pipeline automatically delivers enqueued records to the collector gateway, providing at-least-once delivery even when the collector is temporarily unavailable.
+
+Once records arrive at the gateway, they are persisted via the active storage plugin and become available for aggregation and raw queries. All query operations are scoped to the authenticated tenant derived from the caller's SecurityContext. For ingest, tenant attribution is PDP-authorized at emit time: standard sources emit for their own tenant (derived from SecurityContext); sources explicitly authorized by the PDP to act on behalf of multiple tenants supply the target tenant in the record, which is validated against PDP-returned constraints before the outbox INSERT and re-validated at gateway ingest as a defense-in-depth check.
+
+### 1.2 Architecture Drivers
+
+#### Functional Drivers
+
+| Requirement | Design Response |
+|-------------|-----------------|
+| `cpt-cf-usage-collector-fr-ingestion` | SDK `emit()` persists record to local outbox within the caller's transaction; outbox library background pipeline delivers to gateway |
+| `cpt-cf-usage-collector-fr-idempotency` | Counter records require a non-null idempotency key; `emit()` rejects counter records without one (`EmitError::MissingIdempotencyKey`) before any outbox INSERT; gauge records accept a null key; idempotency key is carried in the outbox payload; storage plugin performs idempotent upsert on non-null keys at delivery |
+| `cpt-cf-usage-collector-fr-delivery-guarantee` | Transactional outbox in source's local DB; outbox library retries with exponential backoff; messages that exhaust retry budget are moved to the dead-letter store and surfaced via monitoring |
+| `cpt-cf-usage-collector-fr-counter-semantics` | `ScopedUsageCollectorClientV1.emit()` rejects counter records with a negative value or a missing idempotency key before inserting the outbox row; delta records are stored per-event; the persistent total for a (tenant, metric) pair is the SUM of all active delta records — see §3.7 Counter Accumulation for the full strategy including plugin-level acceleration options |
+| `cpt-cf-usage-collector-fr-gauge-semantics` | Gauge records carry `metric_kind = gauge`; `emit()` applies no monotonicity validation; storage plugin stores values as-is |
+| `cpt-cf-usage-collector-fr-tenant-attribution` | `emit()` uses `record.tenant_id` as supplied by the caller and validates it against PDP constraints in `EmitAuthorization`: if the PDP returned a tenant `In` constraint, `record.tenant_id` must be a member of the allowed set; if no tenant constraint was returned, `record.tenant_id` must equal `ctx.subject_tenant_id()` (preserving standard single-tenant behavior). The validated `tenant_id` is stamped into the outbox row. Gateway validates tenant attribution on ingest: when `record.tenant_id ≠ ctx.subject_tenant_id()`, the gateway calls the PDP with `context_tenant_id(record.tenant_id)` and action `"emit_on_behalf"` to re-authorize before delegating to the plugin. |
+| `cpt-cf-usage-collector-fr-resource-attribution` | `UsageRecord` carries optional `resource_id` and `resource_type` fields; persisted in outbox payload and storage backend; gateway and plugin pass them through without interpretation |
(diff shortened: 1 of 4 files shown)


modules/system/usage_collector/docs/PRD.md
```diff
@@ -0,0 +1,855 @@
+# PRD — Usage Collector
```
capybutler: @Artifizer commented on previous PR:

> I would emphasize below that we are talking here about Usage Ledger / Metering system, because otherwise there will be an intent to add more and more requirements here (e.g complex data aggregation)

I've restructured PRD to focus on collection and exposure of usage data, leaving processing to consumer modules.

modules/system/usage_collector/docs/DESIGN.md
```diff
@@ -0,0 +1,1014 @@
...
+
+| Entity | Description | Schema |
+|--------|-------------|--------|
+| UsageRecord | A single usage data point with tenant, source, resource, and metric attribution | [usage_record.v1.schema.json](../usage-collector-sdk/schemas/usage_record.v1.schema.json) |
+| MeasuringUnit | Registered schema defining how a usage type is measured | [measuring_unit.v1.schema.json](../usage-collector-sdk/schemas/measuring_unit.v1.schema.json) |
+| BackfillOperation | Metadata and audit trail for a backfill request | [backfill_operation.v1.schema.json](../usage-collector-sdk/schemas/backfill_operation.v1.schema.json) |
+| RateLimitConfig | Per-tenant and per-source rate limit parameters | [rate_limit_config.v1.schema.json](../usage-collector-sdk/schemas/rate_limit_config.v1.schema.json) |
+| RetentionPolicy | Per-type and per-tenant data retention configuration | [retention_policy.v1.schema.json](../usage-collector-sdk/schemas/retention_policy.v1.schema.json) |
+
+**Key attributes of UsageRecord**:
```
capybutler: @Artifizer commented on previous PR:

> I'd suggest to have the following usage entries attributes (here or in DESIGN):
> 
> ```
> - tenant id (mandatory)
> - usage type (mandatory, as GTS type)
> - resource id (mandatory, for global per-tenant usage just use tenant ID)
> - value (mandatory)
> - event time (mandatory, when event happened on the source side)
> - ingest time (mandatory, when event was ingested into UC DB)
> - idempotency key (mandatory)
> - source (optional, this is required for distributed systems to distinguish origin of the usage record)
> - ...
> ```
> 
> so then the unique key would be:
> (tenant_id, source_id, usage_type, resource_id, idempotency_key)

I've specified UsageRecord entity to match those requirements.

nonameffh (changes requested):

modules/system/usage_collector/docs/PRD.md
```diff
@@ -0,0 +1,855 @@
...
+
+**ID**: `cpt-cf-uc-actor-storage-backend`
+
+**Role**: Persists usage records (ClickHouse, PostgreSQL, or external system via adapter).
+
+## 3. Operational Concept & Environment
+
+No module-specific environment constraints beyond project defaults.
+
+## 4. Scope
```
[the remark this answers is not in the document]
capybutler: I've added `cpt-cf-uc-fr-record-metadata` requirement to conver this usecase.

modules/system/usage_collector/docs/PRD.md
```diff
@@ -0,0 +1,855 @@
...
+4. Billing System processes page and requests next page using cursor
+5. Steps 3-4 repeat until all pages retrieved
+6. Billing System aggregates and processes complete record set for rating (with optional per-user breakdown)
+
+**Postconditions**:
+- Billing System has accurate, deduplicated raw usage records for its own aggregation and invoice generation
+- No records missed or duplicated due to concurrent insertions during pagination
+- If user filtering was applied, billing can generate per-user chargeback reports
+
+### UC: Real-Time Quota Enforcement
```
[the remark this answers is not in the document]
nonameffh: I am not against defining the boundaries of this module only by storing usage and postponing the details of integration with the Quota Enforcement module to next stages.
However, since the documentation currently states that Quota Enforcement is a client of this module and reverse integration is not implied, problems arise due to the fact that there is no single point of entry—different implementations of the same mechanism, given that the mechanism itself implies consistency coordination, will lead to hard-to-detect bugs.

From my perspective usage collection should be an entry point for both cases of usage collection with and without quota enforcement. This does not mean that it must be specified and implemented in the first phase, but in this case, from my point of view, it is better to exclude confusing/controversial details from the current documentation.

modules/system/usage_collector/docs/PRD.md
```diff
@@ -0,0 +1,878 @@
...
+
+### 5.2 Metric Semantics
+
+#### Counter Metric Semantics
+
+- [ ] `p1` - **ID**: `cpt-cf-uc-fr-counter-semantics`
+
+The system **MUST** enforce counter semantics (monotonically increasing values), validate increments, compute deltas, and detect/reject counter violations.
+
+**Rationale**: Counters represent cumulative totals (e.g., total API calls); violations indicate data corruption or misconfigured sources.
```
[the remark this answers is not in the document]
capybutler: The PRD requirement is correct for a usage metering system - counter resets cannot be reliably distinguished from data corruption, which would compromise billing accuracy and audit integrity. Sources unable to maintain persistent counters across restarts must use discrete event reporting (one event per measurable unit) or gauges instead.

modules/system/usage_collector/docs/DESIGN.md
```diff
@@ -0,0 +1,1027 @@
...
+
+Usage records are append-only. The system never performs in-place mutation or hard deletion of usage data. Corrections are handled through amendments (status change to deprecated) or backfill operations (status change to archived with replacement records inserted). This ensures a complete audit trail and supports dispute resolution.
+
+#### Source-of-Truth Simplicity
+
+- [ ] `p1` - **ID**: `cpt-cf-uc-principle-source-of-truth`
+
+UC stores and exposes raw usage records only. No aggregation, rating, pricing, or interpretation is performed by the metering ledger. Downstream consumers (billing, monitoring, quota enforcement) own all derived logic and apply their own aggregation rules to raw records.
+
+#### Non-Blocking Emission
```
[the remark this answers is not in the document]
nonameffh: From my perspective it's a bit related to the question I raised about quota enforcement integration as for such operation we'll need blocking mode with the reservation pattern, but I suppose this decision should be postponed until we introduce quota enforcement requirements and understand how it should be integrated with usage collector

nonameffh (approved):

modules/system/usage_collector/docs/PRD.md
```diff
@@ -0,0 +1,877 @@
...
+The system **MUST** maintain 99.95% monthly availability for usage collection endpoints.
+
+**Threshold**: 99.95% uptime per calendar month
+**Rationale**: Usage collection is on the critical path for all billable operations.
+
+#### Ingestion Throughput
+
+- [ ] `p1` - **ID**: `cpt-cf-uc-nfr-throughput`
+
+The system **MUST** support sustained ingestion of at least 600,000 usage records per minute (10,000 events per second) under normal operation.
```
zaslavskii: What if our system ingests more, for example 50k events per seconds, shall we consider sharding in this case?

zaslavskii: Looks like answer is here https://github.com/cyberfabric/cyberfabric-core/pull/619/changes?diff=unified#diff-644fb7b318ecc982063e74437758e75a82fc1bcc0fcb046a9d690ab3bd9d4117R620

capybutler:
Specifications were split into two modules:
- `usage-collector` that focuses on efficient writing of Usage Records
- `usage-query` that focuses on retrieval of aggregated Usage Records

Both modules are extensible via plugins for specific storage backends. I've replaced PostgreSQL in specifications with TimescaleDB as that's what VendorV will be using for usage storage.

modules/system/usage-collector/docs/DESIGN.md
```diff
@@ -0,0 +1,733 @@
...
+                              ▼                   ▼                   ▼
+                     ┌──────────────┐   ┌──────────────┐   ┌──────────────┐
+                     │  ClickHouse  │   │ TimescaleDB  │   │   Custom     │
+                     │   Database   │   │   Database   │   │   Backend    │
+                     └──────────────┘   └──────────────┘   └──────────────┘
+```
+
+#### UC Gateway
+
+**ID**: `cpt-cf-usage-collector-component-gateway`
```
Artifizer: Can we please have checkboxes and priorities everywhere? Otherwise it's hard to see what it the p1 scope and where to put more attention

modules/system/usage-collector/docs/DESIGN.md
```diff
@@ -0,0 +1,733 @@
+# Technical Design — Usage Collector
+
+## 1. Architecture Overview
+
+### 1.1 Architectural Vision
+
+The Usage Collector follows a **plugin-based gateway pattern** aligned with CyberFabric's modkit architecture. The gateway module orchestrates the write path — ingestion, validation, deduplication, and routing — while delegating persistence to pluggable storage backend plugins (ClickHouse, TimescaleDB, custom). Each plugin implements the full write-path interface for its backend: record persistence, deduplication storage, retention enforcement, and failure buffering.
```
Artifizer: My understanding of the original split between **Usage Collector** and **Usage Query** was that it was meant to isolate two different concerns and workloads.

The collector side would focus primarily on ingestion:

* accept events quickly
* persist them durably
* remain stable even if historical storage is under load

The query side would then handle:

* historical storage
* aggregation
* retention policies
* heavier analytical workloads

In the current proposal the collector already owns a pluggable time-series storage backend, which seems to blur that boundary. It starts coupling ingestion reliability with the behavior and performance characteristics of analytical databases.

One pattern that might preserve the original separation more clearly is introducing an intermediate event layer.

For example:

Service
→ submit usage event to a local event manager (durable local write)
→ producer library forwards events
→ central event manager / stream
→ usage collector consumes events
→ writes to historical storage

In that model the collector becomes a consumer of an event stream rather than the first system responsible for durable storage.

Potential advantages:

* ingestion remains fast and isolated from historical DB performance
* GTS allows to control access to these events (so that only Usage Collector would have permissions to read usage events)
* GTS events schema would also allow to enforce the usage payload validation on Event Manager side
* retries and buffering happen naturally in the event layer
* analytical storage can evolve independently and system can even survive in case of downtimes
* easier backfills and reprocessing (Usage Collector will just read new events later)
* clearer responsibility boundaries

The planned event management system would support both delivery mechanisms:
a) POST to event manager directly (for services that do not have it's own DB or that want to get synchronous delivery)
b) WRITE and COMMIT to local service DB with asynchronous delivery of the locally-committed events to the event manager by special 'event producer' library

these 2 options would help to support any kind of services producing usage events

This also aligns with the **original intent** of separating ingestion from historical processing.

modules/system/usage-collector/docs/DESIGN.md
```diff
@@ -0,0 +1,733 @@
...
+
+#### SDK Client
+
+**ID**: `cpt-cf-usage-collector-component-sdk`
+
+Client-side library providing the primary ingestion interface for platform services (`cpt-cf-usage-collector-fr-sdk-retry`). Implements non-blocking, batched emission with configurable count and time thresholds. Buffers events in a bounded in-memory queue with automatic retry using exponential backoff and jitter on rate limit errors, honoring `retry_after_ms` when present. Under buffer exhaustion, drops oldest events and emits loss metrics (`usage_events_dropped_total`, `usage_events_dropped_rate`, `usage_buffer_utilization`) tagged by `usage_type`, `tenant_id`, and `drop_reason`. Alerts trigger when drop rate exceeds configurable thresholds. Provides Rust API (in-process) and gRPC transports.
```
Artifizer: I think in-memory events delivery is a bit risky... .This is one more reason when event-based usage tracking would be probably benefitial

PR #707, opened 2026-02-21
doc(odata)- OData documentation is aligned with implementation

MikeFalcon77:
## Summary by CodeRabbit

* **Documentation**
  * Consolidated and replaced the standalone OData $select guidance with a unified pagination & select guide.
  * Clarified pagination and field-projection rules, dot-notation examples, validation constraints, error mapping, and distinct list vs single-resource guidance.
  * Updated examples and checklist to reflect new page shape and projection helpers.

* **Chores**
  * Updated cross-references and removed deprecated duplicate documentation entries.


docs/modkit_unified_system/07_odata_pagination_select_filter.md
```diff
@@ -134,70 +152,226 @@ impl UserService {
...
+pub fn page_to_projected_json<T: serde::Serialize>(
+    page: &modkit_odata::Page<T>,
+    selected_fields: Option<&[String]>,
+) -> modkit_odata::Page<Value>
 ```
 
+### $select limitations
+
+- Field projection happens at the application layer, not the database layer
+- The `$select` parameter does not affect database query performance
```
aviator5: Actually, it depends on the concrete RDMBS and may affect database query performance. Why do we want to mention this?

docs/modkit_unified_system/07_odata_pagination_select_filter.md
```diff
@@ -134,70 +152,226 @@ impl UserService {
...
 }
 ```
 
-### page_to_projected_json
+### Dot notation behavior
+
+1. **Entire Parent Selection**: If you select `access_control` without dot notation, the entire nested object is included with all its fields.
+2. **Specific Nested Fields**: If you select `access_control.read` and `access_control.write`, only those specific fields are included in the nested object.
+3. **Deep Nesting**: Dot notation works at any depth: `user.profile.name`, `user.profile.settings.notifications`, etc.
+4. **Case Insensitivity**: Both parent and nested field names are case-insensitive: `AccessControl.Read` is equivalent to `access_control.read`.
```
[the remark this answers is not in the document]
MikeFalcon77: Good catch. The example was wrong - the impl uses .to_lowercase() on both sides, so Access_Control.READ matches access_control.read, but AccessControl does not match access_control. No camelCase <=> snake_case conversion happens. Fixed the doc.

docs/modkit_unified_system/07_odata_pagination_select_filter.md
MikeFalcon77: fixed

PR #719, opened 2026-02-23
Add metrics recorder with cardinality limitation, REAL-163

genericaccount-de:
Add metrics recorder with cardinality limitation.

REAL-163

**Note**: current PR contains both metrics mechanism implementation for presentation and comparison discussion.

## Summary by CodeRabbit

* **New Features**
  * OpenTelemetry metrics collection and export added (disabled by default); startup attempts initialization and logs failures without aborting.
  * Example service now records a get_user counter and exposes metrics to the router.

* **Configuration**
  * New metrics subsections in server and quickstart configs with OTLP exporter options and timeout defaults.

* **Shutdown**
  * Graceful shutdown now flushes metrics before tracing.

* **Tests**
  * Added tests validating metric recording and provider behavior.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -280,8 +280,8 @@ tracing-opentelemetry = "0.32"
 tracing-error = "0.2"
 tracing-appender = "0.2"
 tracing-test = "0.2"
-opentelemetry = { version = "0.31", features = ["trace"] }
-opentelemetry_sdk = { version = "0.31", features = ["trace", "rt-tokio"] }
+opentelemetry = { version = "0.31", features = ["trace", "metrics"] }
+opentelemetry_sdk = { version = "0.31", features = ["trace", "rt-tokio", "metrics"] }
 opentelemetry-otlp = { version = "0.31", features = [
     "grpc-tonic",
     "http-proto",
@@ -470,6 +470,9 @@ temp-env = "0.3"
 
 # Additional utilities
 nanoid = "0.4"
+
+# Metrics
+tokio-metrics = { version = "0.4", features = ["rt"] }
 
 # GTS dependencies
 jsonschema = "0.40"
--- a/libs/modkit/src/telemetry/config.rs
+++ b/libs/modkit/src/telemetry/config.rs
@@ -1,6 +1,7 @@
-//! OpenTelemetry tracing configuration types
+//! OpenTelemetry tracing and metrics configuration types
 //!
-//! These types define the configuration structure for OpenTelemetry distributed tracing.
+//! These types define the configuration structure for OpenTelemetry distributed
+//! tracing and metrics.
 
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
@@ -16,6 +17,20 @@ pub struct TracingConfig {
     pub resource: Option<HashMap<String, String>>,
     pub http: Option<HttpOpts>,
     pub logs_correlation: Option<LogsCorrelation>,
+    #[serde(default)]
+    pub metrics: MetricsConfig,
+}
+
+/// Metrics configuration for OpenTelemetry metrics collection
+#[derive(Debug, Clone, Deserialize, Serialize, Default)]
+pub struct MetricsConfig {
+    pub enabled: bool,
+    pub exporter: Option<Exporter>,
+    /// Maximum number of distinct attribute combinations per instrument.
+    /// When the limit is reached, new combinations are folded into an
+    /// overflow data point.  `None` means the SDK default is used.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cardinality_limit: Option<usize>,
 }
 
 #[derive(Debug, Default, Clone, Deserialize, Serialize, PartialEq, Eq, Copy)]
--- a/libs/modkit/src/telemetry/init.rs
+++ b/libs/modkit/src/telemetry/init.rs
@@ -21,6 +21,7 @@ use opentelemetry_sdk::{
     trace::{Sampler, SdkTracerProvider},
 };
 
+use super::config::MetricsConfig;
 #[cfg(feature = "otel")]
 use super::config::TracingConfig;
 #[cfg(feature = "otel")]
@@ -32,7 +33,7 @@ use tonic::metadata::{MetadataKey, MetadataMap, MetadataValue};
 
 /// Build resource with service name and custom attributes
 #[cfg(feature = "otel")]
-fn build_resource(cfg: &TracingConfig) -> Resource {
+pub(crate) fn build_resource(cfg: &TracingConfig) -> Resource {
     let service_name = cfg.service_name.as_deref().unwrap_or("hyperspot");
     let mut attrs = vec![KeyValue::new("service.name", service_name.to_owned())];
 
@@ -102,7 +103,7 @@ fn build_http_exporter(
     if let Some(t) = timeout {
         b = b.with_timeout(t);
     }
-    if let Some(hmap) = build_headers_from_cfg_and_env(cfg) {
+    if let Some(hmap) = build_headers_from_cfg_and_env(cfg.exporter.as_ref()) {
         b = b.with_headers(hmap);
     }
     #[allow(clippy::expect_used)]
@@ -122,7 +123,7 @@ fn build_grpc_exporter(
     if let Some(t) = timeout {
         b = b.with_timeout(t);
     }
-    if let Some(md) = build_metadata_from_cfg_and_env(cfg) {
+    if let Some(md) = build_metadata_from_cfg_and_env(cfg.exporter.as_ref()) {
         b = b.with_metadata(md);
     }
     b.build().context("build OTLP gRPC exporter")
@@ -185,14 +186,14 @@ pub fn init_tracing(
 }
 
 #[cfg(feature = "otel")]
-fn build_headers_from_cfg_and_env(
-    cfg: &TracingConfig,
+pub(crate) fn build_headers_from_cfg_and_env(
+    exporter: Option<&crate::telemetry::config::Exporter>,
 ) -> Option<std::collections::HashMap<String, String>> {
     use std::collections::HashMap;
     let mut out: HashMap<String, String> = HashMap::new();
 
     // From config file
-    if let Some(exp) = &cfg.exporter
+    if let Some(exp) = exporter
         && let Some(hdrs) = &exp.headers
     {
         for (k, v) in hdrs {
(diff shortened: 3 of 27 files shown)


Bechma (changes requested):

MikeFalcon77 (changes requested):

examples/modkit/users-info/users-info/src/domain/metrics.rs
```diff
@@ -0,0 +1,94 @@
+use std::sync::LazyLock;
+
+use opentelemetry::metrics::Counter;
+
+static GET_USER_COUNTER: LazyLock<Counter<u64>> = LazyLock::new(|| {
+    opentelemetry::global::meter("users-info")
+        .u64_counter("users_info.get_user")
```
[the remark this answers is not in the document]
genericaccount-de: It is correct: follows case conventions and separated with dots. It might appear incorrect due to underscore in "users_info", but in our case it is a single logical name

examples/modkit/users-info/users-info/src/domain/metrics.rs
```diff
@@ -0,0 +1,97 @@
+use modkit_macros::domain_model;
+use opentelemetry::metrics::{Counter, Meter};
+
+#[domain_model]
+#[derive(Debug)]
+pub struct UsersMetrics {
```
MikeFalcon77: Let's keep the domain layer free of infrastructure dependencies without introducing generic metric abstractions.

Move OpenTelemetry types out of the domain layer.

1. Define a narrow UsersMetrics trait in the infra layer (e.g. record_get_user()).
2. Move the OTel-based implementation (Counter, Meter, etc.) to infra.
2. Inject the trait into the service and call it there.
3. Do not expose OTel types or infrastructure concerns in domain models.

examples/modkit/users-info/users-info/src/api/rest/routes/mod.rs
```diff
@@ -56,7 +56,13 @@ pub(crate) fn register_routes(
     router = cities::register_city_routes(router, openapi);
     router = addresses::register_address_routes(router, openapi);
 
-    router = router.layer(axum::Extension(services));
+    let metrics = Arc::new(crate::domain::metrics::UsersMetrics::new(
```
MikeFalcon77: Please do not introduce additioanl layer. You need something like that:

let users_metrics: Arc<dyn UsersMetrics> = Arc::new(OtelUsersMetrics::new(&meter));
let services = Arc::new(ConcreteAppServices::new(/* repos..., */ users_metrics));

examples/modkit/users-info/users-info/src/domain/metrics.rs
```diff
@@ -0,0 +1,97 @@
+use modkit_macros::domain_model;
+use opentelemetry::metrics::{Counter, Meter};
```
MikeFalcon77: Domain layer directly depends on `opentelemetry` — an infrastructure crate. Per project DDD-light rules, domain depends only on ports/traits, not on concrete infra. `Counter<u64>` and `Meter` belong in `infra/`, not `domain/`. Also `#[domain_model]` on a struct wrapping an OTel counter is semantically wrong.

Expected pattern:
- `domain/ports/metrics.rs` — `pub trait UsersMetricsPort: Send + Sync { fn record_get_user(&self); }`
- `infra/metrics.rs` — `struct OtelUsersMetrics { counter: Counter<u64> }` + `impl UsersMetricsPort`
- Module bootstrap wires `Arc<dyn UsersMetricsPort>` into Extension
- Handler receives `Arc<dyn UsersMetricsPort>`, not the concrete struct

examples/modkit/users-info/users-info/src/domain/metrics.rs
```diff
@@ -0,0 +1,97 @@
...
+        Self {
+            get_user_counter: meter
+                .u64_counter("users_info.get_user")
+                .with_description("Number of get_user calls")
+                .build(),
+        }
+    }
+
+    pub fn record_get_user(&self) {
+        self.get_user_counter.add(1, &[]);
```
MikeFalcon77: Counter with no attributes can't distinguish success from error or break down by any dimension. For an example that's supposed to show how to add metrics to a module, this is a bit too minimal to be useful — at least show one attribute (e.g. `outcome: ok/error`).

libs/modkit/src/telemetry/init.rs
```diff
@@ -286,6 +291,70 @@ pub fn shutdown_tracing() {
...
 }
 
+/// Gracefully shut down OpenTelemetry metrics.
+/// In opentelemetry 0.31 there is no global `shutdown_meter_provider()`.
+/// Keep a handle to `SdkMeterProvider` in your app state and call `shutdown()`
+/// during graceful shutdown. This function remains a no-op for compatibility.
+#[cfg(feature = "otel")]
+pub fn shutdown_metrics() {
+    tracing::info!("Metrics shutdown: no-op (keep a provider handle to call `shutdown()`).");
+}
```
MikeFalcon77: Real problem: metrics won't be flushed on shutdown. The `SdkMeterProvider` constructed in `do_init_metrics_provider` is passed to `global::set_meter_provider` and then dropped — no handle is retained anywhere. When this gets called at shutdown nothing gets flushed, and anything buffered in the OTLP exporter is lost.

libs/modkit/src/telemetry/init.rs
```diff
@@ -286,6 +291,70 @@ pub fn shutdown_tracing() {
...
+///
+/// # Errors
+/// Returns an error if `TracingConfig::metrics` is `None` or its `enabled`
+/// field is `false`.
+#[cfg(feature = "otel")]
+#[allow(clippy::needless_pass_by_value)]
+pub fn init_metrics(
+    cfg: &TracingConfig,
+    provider: opentelemetry_sdk::metrics::SdkMeterProvider,
+) -> anyhow::Result<opentelemetry::metrics::Meter> {
```
MikeFalcon77: `init_metrics` returns `anyhow::Result<Meter>` but the only caller (`do_init_metrics_provider`) discards it immediately with `?`. All real callers use `opentelemetry::global::meter()` after init. Return type should be `anyhow::Result<()>`.

libs/modkit/src/telemetry/init.rs
```diff
@@ -286,6 +291,70 @@ pub fn shutdown_tracing() {
...
+    tracing::info!("OpenTelemetry metrics initialized successfully");
+    Ok(meter)
+}
+
+/// Initialize OpenTelemetry metrics (no-op when otel feature is disabled).
+///
+/// # Errors
+/// This function always returns an error when the otel feature is disabled.
+#[cfg(not(feature = "otel"))]
+pub fn init_metrics(_cfg: &serde_json::Value, _provider: ()) -> anyhow::Result<()> {
```
MikeFalcon77: The non-otel stub signature (`&serde_json::Value, ()`) doesn't match the otel version (`&TracingConfig, SdkMeterProvider`). Any caller that wants to compile on both feature paths still needs a `#[cfg]` guard, so this stub provides no value. Either match the signature or drop the stub and guard all call sites with `#[cfg(feature = "otel")]`.

libs/modkit/src/telemetry/config.rs
```diff
@@ -16,6 +17,19 @@ pub struct TracingConfig {
...
+/// Metrics configuration for OpenTelemetry metrics collection
+#[derive(Debug, Clone, Deserialize, Serialize, Default)]
+pub struct MetricsConfig {
+    pub enabled: bool,
+    pub exporter: Option<Exporter>,
+    /// Maximum number of distinct attribute combinations per instrument.
+    /// When the limit is reached, new combinations are folded into an
+    /// overflow data point.  `None` means the SDK default is used.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cardinality_limit: Option<u32>,
```
MikeFalcon77: Cast to `usize` in `metrics_provider.rs` (`limit as usize`). Store as `usize` directly to avoid the cast and match the type the SDK expects.

Cargo.toml
```diff
@@ -471,6 +471,11 @@ temp-env = "0.3"
 # Additional utilities
 nanoid = "0.4"
 
+# Metrics
+metrics = "0.24"
```
MikeFalcon77: `metrics` and `metrics-exporter-prometheus` are not used by any file in this PR. `tokio-metrics` added to `libs/modkit/Cargo.toml` as dev-dep is also unused. Looks like leftovers from an earlier approach — should be removed.

Bechma (changes requested):

examples/modkit/users-info/users-info/src/domain/metrics.rs
MikeFalcon77: Please remove this file. We don't need backward compatibility in examples.

examples/modkit/users-info/users-info/src/api/rest/routes/mod.rs
```diff
@@ -56,7 +56,15 @@ pub(crate) fn register_routes(
     router = cities::register_city_routes(router, openapi);
     router = addresses::register_address_routes(router, openapi);
 
-    router = router.layer(axum::Extension(services));
+    let scope = opentelemetry::InstrumentationScope::builder("users-info".to_owned()).build();
+    let metrics: Arc<dyn crate::domain::ports::UsersMetricsPort> =
```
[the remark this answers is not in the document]
genericaccount-de: Yes, but this was the point of @MikeFalcon77 's previous comment - to build additional abstraction between register_routes and concrete MetricsMeter implementation and use UsersMetricsPort instead - if I got it correctly. Like it is done for AuditPort

examples/modkit/users-info/users-info/src/domain/metrics.rs
MikeFalcon77: My fault. I should leave more specific comment. 

The port/adapter split for metrics is the right call  - trait in `domain/ports`, otel impl in `infra/`.

One thing worth changing: instead of adding `Arc<dyn UsersMetricsPort>` as a separate `Extension`, put it in `ConcreteAppServices`. It's already in Extension everywhere, so handlers can call `svc.metrics.record_early_fail()` for pre-domain errors, and the domain service gets the same `Arc` injected at construction time in the module bootstrap. 

One Extension instead of two, metrics are accessible from both layers.

MikeFalcon77 (changes requested):

MikeFalcon77 (approved):

Bechma (approved):

Issue #460, opened 2026-02-05
Add API prefix for api_gateway for openapi.json


MikeFalcon77:
For now, we have an issue:

1. The Kubernetes chart adds a prefix to the inbound API, for example: /chat.
2. The OpenAPI spec URL is embedded in the OpenAPI UI JavaScript as /openapi.json.

As a result, it does not work: the Kubernetes Ingress does not know how to redirect this browser request.

We need to introduce a configurable API prefix parameter.

PR #783, opened 2026-02-27
460: Added prefix to path

dominic1988-lgtm:
## Summary by CodeRabbit

* **New Features**
  * API Gateway supports a configurable prefix path so routes, docs, and OpenAPI can be served under a specified prefix (e.g., /cf).
* **Behavioral Fixes**
  * Route matching, MIME validation, license checks, authentication, and rate limits now honor the configured prefix.
* **Tests**
  * Expanded tests for prefixed routes, docs, middleware order, CORS, and path-parameter scenarios.
* **Chore**
  * Default OpenAPI URL updated to the prefix-aware path; minor formatting and test adjustments.

### DIFF
--- a/config/quickstart.yaml
+++ b/config/quickstart.yaml
@@ -90,7 +90,7 @@ modules:
       bind_addr: "127.0.0.1:8087"
       enable_docs: true
       cors_enabled: false
-
+      prefix_path: "/cf"
       # OpenAPI Documentation Configuration
       openapi:
         title: "HyperSpot API"
--- a/modules/system/api-gateway/src/module.rs
+++ b/modules/system/api-gateway/src/module.rs
@@ -77,6 +77,19 @@ impl Default for ApiGateway {
 }
 
 impl ApiGateway {
+    fn apply_prefix_nesting(mut router: Router, prefix: &str) -> Router {
+        if prefix.is_empty() {
+            return router;
+        }
+
+        let top = Router::new()
+            .route("/health", get(web::health_check))
+            .route("/healthz", get(|| async { "ok" }));
+
+        router = Router::new().nest(prefix, router);
+        top.merge(router)
+    }
+
     /// Create a new `ApiGateway` instance with the given configuration
     #[must_use]
     pub fn new(config: ApiGatewayConfig) -> Self {
@@ -125,11 +138,13 @@ impl ApiGateway {
         // Always mark built-in health check routes as public
         public_routes.insert((Method::GET, "/health".to_owned()));
         public_routes.insert((Method::GET, "/healthz".to_owned()));
+
         public_routes.insert((Method::GET, "/docs".to_owned()));
         public_routes.insert((Method::GET, "/openapi.json".to_owned()));
 
         for spec in &self.openapi_registry.operation_specs {
             let spec = spec.value();
+
             let route_key = (spec.method.clone(), spec.path.clone());
 
             if spec.authenticated {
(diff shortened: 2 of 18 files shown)


dominic1988-lgtm:
https://github.com/cyberfabric/cyberfabric-core/issues/460

modules/system/api-gateway/src/module.rs
```diff
@@ -158,6 +160,17 @@ impl ApiGateway {
         Ok(route_policy)
     }
 
+    fn normalize_prefix_path(raw: &str) -> String {
+        let prefix = raw.trim().trim_end_matches('/');
+        if prefix.is_empty() || prefix == "/" {
```
Bechma: prefix can't be "/" as / was trimmed before

Bechma (approved):
LGTM 👍

MikeFalcon77 (changes requested):
.

modules/system/api-gateway/src/web.rs
```diff
@@ -28,9 +28,14 @@ pub async fn health_check() -> Json<Value> {
 }
 
 #[cfg(not(feature = "embed_elements"))]
-pub async fn serve_docs() -> Html<&'static str> {
+pub fn serve_docs(prefix_path: &str) -> Html<String> {
+    let openapi_url = if prefix_path.is_empty() {
+        String::from("/openapi.json")
+    } else {
+        format!("{prefix_path}/openapi.json")
```
MikeFalcon77: `prefix_path` is interpolated into an HTML attribute without any escaping. If the value comes from an env var or remote config store, `"><script>alert(1)</script>` injects into the served page.

Config is not user input, but the risk is real if the config surface is wide. At minimum validate that `prefix_path` matches `^[a-zA-Z0-9/_-]*$` before using it in a template.

modules/system/api-gateway/tests/auth_middleware.rs
```diff
@@ -176,9 +176,9 @@ async fn test_auth_disabled_mode() {
     api_gateway.init(&api_ctx).await.expect("Failed to init");
 
     // Register test module
-    let router = Router::new();
+    let mut router = Router::new();
```
MikeFalcon77: This change (and all the similar ones in this PR) replaces idiomatic Rust variable shadowing with `let mut` + reassignment. `let mut` signals that a value mutates in place — that's not what's happening here. The original `let router = ...` / `let router = ...` shadow pattern is correct and should be kept where the new prefix logic doesn't require it.

modules/system/api-gateway/tests/auth_middleware.rs
```diff
@@ -473,6 +473,74 @@ async fn test_route_pattern_matching_with_path_params() {
...
+        )
+        .await
+        .expect("Request failed");
+
+    assert_eq!(
+        response.status(),
+        StatusCode::OK,
+        "Route with different path parameter value should also be accessible"
+    );
+}
```
MikeFalcon77: The test only confirms that prefixed paths return 200. It does not prove that the prefix is actually enforced — a router that ignores the prefix entirely would also pass.

Add a negative case: send a request to `/tests/v1/api/users/123` (without `/cf`) and assert `StatusCode::NOT_FOUND`. Same issue in `rejects_non_base_feature_requirement_with_prefix` and `real_middlewares_observe_documented_order_with_prefix`.

dominic1988-lgtm: Updated

modules/system/api-gateway/src/web.rs
```diff
@@ -28,9 +28,14 @@ pub async fn health_check() -> Json<Value> {
 }
 
 #[cfg(not(feature = "embed_elements"))]
-pub async fn serve_docs() -> Html<&'static str> {
+pub fn serve_docs(prefix_path: &str) -> Html<String> {
```
MikeFalcon77: Why is it sync? It must stay async.

modules/system/api-gateway/src/web.rs
```diff
@@ -40,27 +45,41 @@ pub async fn serve_docs() -> Html<&'static str> {
...
+  <elements-api apiDescriptionUrl="{openapi_url}" router="hash" layout="sidebar"></elements-api>
 </body>
 </html>"#,
-    )
+    ))
 }
 
 #[cfg(feature = "embed_elements")]
-pub async fn serve_docs() -> Html<&'static str> {
+pub fn serve_docs(prefix_path: &str) -> Html<String> {
```
MikeFalcon77: What is a purpose of this change? It must be async.

modules/system/api-gateway/src/web.rs
dominic1988-lgtm: it is sync as we are initializing the html_doc before sending it down to /docs

config/quickstart.yaml
```diff
@@ -90,7 +90,7 @@ modules:
       bind_addr: "127.0.0.1:8087"
       enable_docs: true
       cors_enabled: false
-
+      prefix_path: "/cf"
```
Artifizer: Can we just comment it out? Because all our instructions and examples will likely break
Also, would you please double check that the server prints the right path in the logs when it starts?
Also please ensure /docs work well after the change

Artifizer (changes requested):

PR #810, opened 2026-02-28
Add separate profiling server with CPU pprof endpoint, REAL-163

genericaccount-de:
Add separate profiling server with CPU pprof endpoint.

REAL-163

## Summary by CodeRabbit

* **New Features**
  * On-demand CPU profiling for performance analysis (opt-in, disabled by default)
  * Local profiling server with bearer-token support, default bind 127.0.0.1:6060
  * Two endpoints: CPU profile export (pprof) and flamegraph visualization
  * Configurable sampling frequency and collection duration; profiling can be enabled via a feature toggle (Unix-only)

* **Documentation**
  * Added a detailed profiling guide covering configuration, auth, usage, and deployment notes

### DIFF
--- a/libs/modkit/src/telemetry/mod.rs
+++ b/libs/modkit/src/telemetry/mod.rs
@@ -1,12 +1,18 @@
-//! Telemetry utilities for OpenTelemetry integration
+//! Telemetry utilities for OpenTelemetry integration and CPU profiling
 //!
 //! This module provides utilities for setting up and configuring
-//! OpenTelemetry tracing layers for distributed tracing.
+//! OpenTelemetry tracing layers for distributed tracing, and
+//! on-demand CPU profiling via pprof-compatible endpoints.
 
 pub mod config;
 pub mod init;
 pub mod throttled_log;
 
-pub use config::{Exporter, HttpOpts, LogsCorrelation, Propagation, Sampler, TracingConfig};
+pub use config::{
+    Exporter, HttpOpts, LogsCorrelation, ProfilingConfig, Propagation, Sampler, TracingConfig,
+};
+
+#[cfg(all(unix, feature = "profiling"))]
+pub mod profiling;
 pub use init::{init_tracing, shutdown_tracing};
 pub use throttled_log::ThrottledLog;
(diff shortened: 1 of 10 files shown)


libs/modkit/src/telemetry/profiling.rs
```diff
@@ -0,0 +1,373 @@
...
+            .unwrap_or(DEFAULT_FREQUENCY)
+            .clamp(1, MAX_FREQUENCY)
+    }
+}
+
+// ---------------------------------------------------------------------------
+// ${ENV_VAR} expansion for auth_token
+// ---------------------------------------------------------------------------
+
+fn expand_env_vars(input: &str) -> Result<String> {
```
MikeFalcon77: we have the same function in libs/modkit-db/src/options.rs
LEt's move it into the modkit-utils

libs/modkit/src/telemetry/profiling.rs
```diff
@@ -0,0 +1,373 @@
...
+        result.replace_range(start..=start + end, &value);
+    }
+    Ok(result)
+}
+
+// ---------------------------------------------------------------------------
+// Auth middleware
+// ---------------------------------------------------------------------------
+
+async fn auth_middleware(
```
MikeFalcon77: Auth middleware looks strange for profiling

libs/modkit/src/telemetry/mod.rs
```diff
@@ -1,12 +1,18 @@
...
 pub mod config;
 pub mod init;
 pub mod throttled_log;
 
-pub use config::{Exporter, HttpOpts, LogsCorrelation, Propagation, Sampler, TracingConfig};
+pub use config::{
+    Exporter, HttpOpts, LogsCorrelation, ProfilingConfig, Propagation, Sampler, TracingConfig,
+};
+
+#[cfg(all(unix, feature = "profiling"))]
```
Bechma: only unix? what happens on windows? In docs say it is optimized for linux, but I undertand that is supported

Bechma (changes requested):

genericaccount-de:
obsolete

PR #822, opened 2026-03-01
fix: `contains_uuid` misses UUID-as-String scope values

lizreu:
## Summary

`AccessScope::contains_uuid` used strict `PartialEq` (via `contains_value`),
so a UUID stored as `ScopeValue::String("550e8400-...")` would never match even
though `uuid_values` / `as_uuid` already parsed such strings. This inconsistency
could silently produce incorrect authorization decisions. The fix rewrites
`contains_uuid` to use `as_uuid()` for comparison, matching both `Uuid` and
`String` variants.

## Changes

- `libs/modkit-security/src/access_scope.rs` — `AccessScope::contains_uuid`
  now iterates filter values with `v.as_uuid() == Some(id)` instead of
  delegating to `contains_value`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UUID detection in access scope validation to correctly handle both UUID and string representations, ensuring accurate matching across all constraint types.

* **Tests**
  * Added unit tests to verify UUID string variant matching and prevent false positives with invalid string formats.


Bechma (changes requested):
Could you fix the CI?

Bechma (approved):

MikeFalcon77 (approved):

PR #823, opened 2026-03-01
fix(modkit-db): warn on malformed database config

lizreu:
## Summary

`DbManager` silently discarded `serde_json` deserialization errors via `.ok()`,
turning a config typo (e.g. `server` instead of `servers` under
`#[serde(deny_unknown_fields)]`) into a confusing downstream error
("Referenced server 'main' not found") with no indication of the real cause.
Additionally, `build_for_module` collapsed both "key absent" and "key present
but malformed" into `None`, causing the misleading `debug!("Module has no
database configuration; skipping")` message to fire even after a deserialization
failure had already been handled.

This PR fixes both issues by replacing `.ok()` call sites with explicit match
arms that emit `tracing::warn!` on failure, and restructuring `build_for_module`
into two distinct early-return paths so each diagnostic log fires only for its
own case.

## Changes

- `libs/modkit-db/src/manager.rs`: warn on malformed global `database` config in `from_figment()`
- `libs/modkit-db/src/manager.rs`: warn on malformed module `database` config in `build_for_module()`, including the module name in the log field
- `libs/modkit-db/src/manager.rs`: restructure `build_for_module` so `debug!` fires only when the key is genuinely absent, and `warn!` + early return fires only on deserialization failure

### DIFF
--- a/libs/modkit-db/src/manager.rs
+++ b/libs/modkit-db/src/manager.rs
@@ -96,31 +96,25 @@ impl DbManager {
             .extract()
             .unwrap_or_else(|_| serde_json::Value::Object(serde_json::Map::new()));
 
-        let module_cfg: Option<DbConnConfig> = match module_data
+        let Some(db_value) = module_data
             .get("modules")
             .and_then(|modules| modules.get(module))
             .and_then(|m| m.get("database"))
-        {
-            None => None,
-            Some(db) => match serde_json::from_value(db.clone()) {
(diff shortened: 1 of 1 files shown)


MikeFalcon77:
Hi, please sign-off you change (DCO check is failed)

MikeFalcon77 (approved):

Bechma (approved):

MikeFalcon77 (approved):

PR #824, opened 2026-03-01
Fix expand_env_vars double-expansion when env values contain ${...}

lizreu:
## Summary

Fixes a correctness bug in `expand_env_vars` where environment variable values containing `${...}` substrings were incorrectly re-expanded. The previous implementation used `captures_iter` over the original input but `String::replace` on a mutating result, so substitutions could introduce new placeholders that got expanded again. This fix uses `Regex::replace_all` with a closure for single-pass expansion.

## Changes

- `libs/modkit-db/src/options.rs`: Replace iterative expansion with single-pass `replace_all`; use `LazyLock` for regex; add regression test

### DIFF
--- a/libs/modkit-db/src/options.rs
+++ b/libs/modkit-db/src/options.rs
@@ -695,15 +695,19 @@ fn parse_sqlite_path_from_dsn(dsn: &str) -> Result<std::path::PathBuf> {
 ///
 /// Uses single-pass replacement so values containing `${...}` are not re-expanded.
 fn expand_env_vars(input: &str) -> Result<String> {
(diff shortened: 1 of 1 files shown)


MikeFalcon77:
Thanks for the work on this!

I incorporated the fixes from this PR into a new PR together with some additional changes, so I’m going to close this one to avoid duplication.

The new PR: [<link to your PR>](https://github.com/cyberfabric/cyberfabric-core/pull/966)

Thanks again for the investigation and the proposed fixes.

MikeFalcon77:
See comment: https://github.com/cyberfabric/cyberfabric-core/pull/824#issuecomment-4023121081

PR #825, opened 2026-03-01
fix(grpc): correct backoff to exponential and add jitter

lizreu:
## Summary

`connect_with_retry` and `call_with_retry` in `modkit-transport-grpc` computed backoff as
`base * attempt` (linear growth), despite docs and comments advertising exponential backoff.
Neither function added jitter, creating a thundering-herd risk when many clients retry
simultaneously. This PR fixes both issues to match the behaviour already present in
`modkit-http`.

## Changes

- `libs/modkit-transport-grpc/Cargo.toml` — add `rand` workspace dependency
- `libs/modkit-transport-grpc/src/client.rs` — `connect_with_retry`: change backoff formula
  from `base * attempt` to `base * 2^(attempt-1)`, capped at `max_backoff`, then add 0–25 %
  random jitter; update doc comment to match
- `libs/modkit-transport-grpc/src/rpc_retry.rs` — `call_with_retry`: same formula change and
  jitter; update module doc, `base_backoff` field doc, and inline comment

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retry behavior for connections and RPCs: switched to jittered exponential backoff with strict max cap to reduce repeated failures and avoid unbounded delays.
* **Documentation**
  * Clarified configuration docs for retry-related settings (max retries, base and max backoff) to reflect shared usage across connection and RPC retries.
* **Tests**
  * Added unit tests validating backoff growth, capping, jitter, and robustness for large attempt counts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1431,6 +1431,7 @@ version = "0.2.14"
 dependencies = [
  "anyhow",
  "cf-modkit-security",
+ "rand 0.9.2",
  "tokio",
  "tokio-stream",
  "tokio-util",
--- a/libs/modkit-transport-grpc/Cargo.toml
+++ b/libs/modkit-transport-grpc/Cargo.toml
@@ -25,3 +25,4 @@ tracing = { workspace = true }
 tokio = { workspace = true }
 tokio-util = { workspace = true }
 tokio-stream = { workspace = true }
+rand = { workspace = true }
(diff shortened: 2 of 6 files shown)


Bechma (approved):

Bechma (changes requested):
could you deal with the comments? also looks like the CI is not passing

Bechma (changes requested):

MikeFalcon77 (approved):

PR #826, opened 2026-03-01
fix(modkit): log forwarder panics and drain timeouts in wait_forwarder

lizreu:
## Summary

`wait_forwarder` in `libs/modkit/src/backends/local.rs` discarded both the
`JoinError` produced when a log-forwarder task panics and the timeout sentinel,
making failures completely invisible in production logs. This PR matches on all
three arms of the `timeout` result and emits a `warn!` for each failure path.

## Changes

- `libs/modkit/src/backends/local.rs`: replace `_ = tokio::time::timeout(…, h).await` with a full `match` that logs forwarder panics and drain timeouts.

### DIFF
--- a/libs/modkit/src/backends/local.rs
+++ b/libs/modkit/src/backends/local.rs
@@ -139,15 +139,30 @@ async fn stop_child_with_grace(
 }
 
 /// Wait for a log forwarder task to finish with timeout.
-async fn wait_forwarder(handle: Option<JoinHandle<()>>) {
+async fn wait_forwarder(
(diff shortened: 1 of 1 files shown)


MikeFalcon77:
Please sign-off your change (DCO check is failed)

Bechma (changes requested):
Could you apply the comments from copilot? It will help in logging

Bechma (approved):

MikeFalcon77 (approved):

MikeFalcon77:
Rest of comments will be addressed at the next PRs

PR #828, opened 2026-03-01
chore(modkit-db): remove orphan errors.rs (never compiled)

lizreu:
## Summary

Remove the orphan `libs/modkit-db/src/errors.rs` file. It was never declared as a module (`mod errors;`) in the crate, so it was never compiled. The functions inside had zero callers. This cleans up dead code.

## Changes

- Remove `libs/modkit-db/src/errors.rs`


Bechma (approved):

MikeFalcon77:
But how will we check for constraint violation errors if someone needs that?

PR #829, opened 2026-03-01
fix(modkit-utils): add Deserialize impl for SecretString behind serde feature

lizreu:
## Summary

`SecretString` had no `Deserialize` impl, forcing callers to deserialize a plain `String` first and then wrap it — leaving a short-lived but unprotected `String` on the heap that `zeroize` never sees. This PR adds a `Deserialize` impl gated behind a new `serde` feature flag so the conversion happens in one step with no intermediate allocation.

## Changes

- `libs/modkit-utils/Cargo.toml`: expose a standalone `serde` feature (`dep:serde`); the `humantime-serde` feature now implies the crate-level `serde` feature (rather than only `dep:serde`) so the `SecretString` `Deserialize` impl is reachable when `humantime-serde` is enabled.
- `libs/modkit-utils/src/secret_string.rs`: implement `serde::Deserialize` for `SecretString` under `#[cfg(feature = "serde")]` using fully-qualified syntax (`<String as serde::Deserialize>::deserialize`) to avoid requiring the trait to be in scope; add a unit test exercising JSON deserialization.

## Summary by CodeRabbit

* **New Features**
  * Optional serialization/deserialization support added for library types; enabling the option allows sensitive string values to be serialized/deserialized safely.

* **Tests**
  * Added a unit test to validate deserialization of the sensitive string type when serialization support is enabled.

### DIFF
--- a/libs/modkit-utils/Cargo.toml
+++ b/libs/modkit-utils/Cargo.toml
@@ -19,7 +19,7 @@ workspace = true
 
 [features]
 serde = ["dep:serde"]
-humantime-serde = ["dep:humantime", "dep:serde"]
+humantime-serde = ["serde", "dep:humantime"]
 
 [dependencies]
 serde = { workspace = true, optional = true }
--- a/libs/modkit-utils/Cargo.toml
+++ b/libs/modkit-utils/Cargo.toml
@@ -18,6 +18,7 @@ name = "modkit_utils"
 workspace = true
 
 [features]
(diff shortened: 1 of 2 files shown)


Bechma (changes requested):
Could you deal with bot comments, I think they made a point with serde feature

Bechma (approved):

MikeFalcon77 (approved):

PR #852, opened 2026-03-03
feat: oagw pingora-based proxy

mattgarmon:
## Summary by CodeRabbit

* **New Features**
  * In-process Pingora proxy/bridge with DNS-aware discovery, per-upstream load‑balancing, session bridge, PingoraEndpointSelector, max-body-size and allow-http-upstream options.
  * Public helpers: proxy error-source extractor and test helpers (MockUpstream, backend selector injection).

* **Bug Fixes**
  * Load‑balancer cache invalidation on upstream/route CRUD and single-key rate‑limit removal.

* **Improvements**
  * Explicit WebSocket rejection, earlier TLS provider initialization, stricter endpoint validation, and refined header sanitization.

* **Tests & Docs**
  * Expanded unit/integration/e2e tests and new scenarios for unreachable backends and cache invalidation.

### DIFF
--- a/modules/system/oagw/oagw/src/domain/services/management.rs
+++ b/modules/system/oagw/oagw/src/domain/services/management.rs
@@ -1,12 +1,16 @@
 use std::sync::Arc;
 
 use super::ControlPlaneService;
+use std::net::IpAddr;
+
 use crate::domain::error::DomainError;
 use crate::domain::model::{
-    CreateRouteRequest, CreateUpstreamRequest, ListQuery, Route, UpdateRouteRequest,
+    CreateRouteRequest, CreateUpstreamRequest, Endpoint, ListQuery, Route, UpdateRouteRequest,
     UpdateUpstreamRequest, Upstream,
 };
 use crate::domain::repo::{RouteRepository, UpstreamRepository};
+
+use async_trait::async_trait;
 use modkit_macros::domain_model;
 use modkit_security::SecurityContext;
 use uuid::Uuid;
@@ -26,6 +30,74 @@ impl ControlPlaneServiceImpl {
     ) -> Self {
         Self { upstreams, routes }
     }
+}
+
+/// Validate the endpoint list for a server configuration.
+///
+/// Rules:
+/// - At least one endpoint is required.
+/// - All endpoints must use either IP addresses or hostnames — no mixing.
+/// - All endpoints must share the same scheme (upstream-level invariant).
+fn validate_endpoints(endpoints: &[Endpoint]) -> Result<(), DomainError> {
+    if endpoints.is_empty() {
+        return Err(DomainError::validation(
+            "server must have at least one endpoint",
+        ));
+    }
+
+    // TODO(hardening): add configurable SSRF deny-list for private IPv4 ranges
+    // (loopback, RFC 1918, link-local, 169.254.169.254 metadata). Should be
+    // opt-in (many deployments legitimately proxy to internal services) and also
+    // enforced at DNS resolution time in DnsDiscovery::resolve() to cover
+    // hostnames that resolve to private IPs.
+
+    // IPv6 endpoints are not yet supported — reject early with a clear message.
+    // Enabling IPv6 requires SSRF protections (deny-lists for link-local, private
+    // ranges, IPv4-mapped addresses).
+    for (i, ep) in endpoints.iter().enumerate() {
+        if strip_brackets(&ep.host)
+            .parse::<std::net::Ipv6Addr>()
+            .is_ok()
+        {
+            return Err(DomainError::validation(format!(
+                "endpoint[{i}] uses IPv6 address '{}'; IPv6 endpoints are not yet supported",
+                ep.host
+            )));
+        }
+    }
+
+    // Check all-IP vs all-hostname consistency.
+    let ip_count = endpoints
+        .iter()
+        .filter(|ep| strip_brackets(&ep.host).parse::<IpAddr>().is_ok())
+        .count();
+    if ip_count != 0 && ip_count != endpoints.len() {
+        return Err(DomainError::validation(
+            "all endpoints must use either IP addresses or hostnames; mixed configurations are not allowed",
+        ));
+    }
+
+    // Enforce identical scheme and port across the pool.
+    if endpoints.len() > 1 {
+        let first_scheme = &endpoints[0].scheme;
+        let first_port = endpoints[0].port;
+        for (i, ep) in endpoints.iter().enumerate().skip(1) {
+            if ep.scheme != *first_scheme {
+                return Err(DomainError::validation(format!(
+                    "endpoint[{i}] scheme {:?} differs from endpoint[0] scheme {:?}; all endpoints must share the same scheme",
+                    ep.scheme, first_scheme
+                )));
+            }
+            if ep.port != first_port {
+                return Err(DomainError::validation(format!(
+                    "endpoint[{i}] port {} differs from endpoint[0] port {}; all endpoints must share the same port",
+                    ep.port, first_port
+                )));
+            }
+        }
+    }
+
+    Ok(())
 }
 
 /// Maximum length for an upstream alias.
@@ -52,6 +124,14 @@ fn validate_alias(alias: &str) -> Result<(), DomainError> {
     Ok(())
 }
 
+/// Strip surrounding `[` and `]` from a host string so that bracketed IPv6
+/// literals (e.g. `[2001:db8::1]`) can be parsed by `Ipv6Addr` / `IpAddr`.
+fn strip_brackets(host: &str) -> &str {
+    host.strip_prefix('[')
+        .and_then(|s| s.strip_suffix(']'))
+        .unwrap_or(host)
+}
+
 /// Generate an alias from the upstream's server endpoints.
 /// Single endpoint: host (standard port omitted) or host:port.
 fn generate_alias(upstream: &Upstream) -> String {
@@ -63,7 +143,7 @@ fn generate_alias(upstream: &Upstream) -> String {
     endpoints[0].alias_contribution()
 }
 
-#[async_trait::async_trait]
+#[async_trait]
 impl ControlPlaneService for ControlPlaneServiceImpl {
     // -- Upstream CRUD --
 
@@ -72,6 +152,8 @@ impl ControlPlaneService for ControlPlaneServiceImpl {
         ctx: &SecurityContext,
         req: CreateUpstreamRequest,
     ) -> Result<Upstream, DomainError> {
+        validate_endpoints(&req.server.endpoints)?;
+
         let tenant_id = ctx.subject_tenant_id();
         let id = Uuid::new_v4();
 
@@ -139,6 +221,7 @@ impl ControlPlaneService for ControlPlaneServiceImpl {
 
         // Apply partial update.
         if let Some(server) = req.server {
+            validate_endpoints(&server.endpoints)?;
             existing.server = server;
         }
         if let Some(protocol) = req.protocol {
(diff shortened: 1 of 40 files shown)


modules/system/oagw/oagw/src/infra/proxy/pingora_proxy.rs
```diff
@@ -0,0 +1,847 @@
...
+    }
+
+    /// Resolve endpoints to `Backend`s and rebuild the reverse-lookup map.
+    fn resolve(&self) -> (BTreeSet<Backend>, HashMap<String, Endpoint>) {
+        let mut backends = BTreeSet::new();
+        let mut map = HashMap::with_capacity(self.endpoints.len());
+
+        for ep in &self.endpoints {
+            let addr_str = format!("{}:{}", ep.host, ep.port);
+
```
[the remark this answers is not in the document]
mattgarmon: Added early detection for ipv6 addresses, will be supported in later phase

modules/system/oagw/oagw/src/infra/proxy/session_bridge.rs
```diff
@@ -0,0 +1,539 @@
...
+    let (status, headers, body_offset) = loop {
+        let mut tmp = [0u8; 4096];
+        let n = io
+            .read(&mut tmp)
+            .await
+            .context("failed to read response from proxy")?;
+        if n == 0 {
+            anyhow::bail!("proxy closed connection before sending response headers");
+        }
+        buf.extend_from_slice(&tmp[..n]);
```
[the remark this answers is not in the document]
mattgarmon: Thanks, added limit

mattgarmon:
@coderabbitai resume

modules/system/oagw/oagw/tests/e2e_http2_test.rs
```diff
@@ -0,0 +1,478 @@
...
+                let io = TokioIo::new(tls_stream);
+
+                let state_svc = state.clone();
+                let service = service_fn(move |req: Request<Incoming>| {
+                    let state = state_svc.clone();
+                    async move {
+                        let method = req.method().to_string();
+                        let uri = req.uri().to_string();
+                        let version = req.version();
+                        let headers: Vec<(String, String)> = req
```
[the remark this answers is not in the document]
mattgarmon: It's okay for test

modules/system/oagw/oagw/src/infra/proxy/pingora_proxy.rs
```diff
@@ -0,0 +1,987 @@
...
+    /// `try_from_iter`). Called from `discover()` which runs in Pingora's
+    /// background service context.
+    fn resolve(&self) -> (BTreeSet<Backend>, HashMap<String, Endpoint>) {
+        let mut backends = BTreeSet::new();
+        let mut map = HashMap::with_capacity(self.endpoints.len());
+
+        for ep in &self.endpoints {
+            let addr_str = format!("{}:{}", ep.host, ep.port);
+
+            match addr_str.as_str().to_socket_addrs() {
```
MikeFalcon77: **[P1 / correctness]** `ToSocketAddrs` is a **blocking** syscall. `discover()` is `async` and runs inside a `tokio::spawn`'d task (line 240). Blocking DNS on a Tokio worker thread stalls the entire worker — under load (many upstreams refreshing simultaneously, slow DNS) this can cascade into a latency spike across unrelated tasks sharing the same worker.

```rust
// reproduce: add a 2-second DNS delay (e.g. via /etc/hosts + iptables DROP to a DNS server)
// then observe all tasks on the same tokio worker stall for 2 s
```

**Fix:** wrap in `tokio::task::spawn_blocking`:
```rust
let addr_str_clone = addr_str.clone();
let resolved = tokio::task::spawn_blocking(move || {
    addr_str_clone.as_str().to_socket_addrs()
}).await.unwrap();
```
Or use `tokio::net::lookup_host` directly.

**Verify:** `tokio::runtime::Builder::new_current_thread()` test with an unresolvable host — blocking variant deadlocks, async variant returns an error.

modules/system/oagw/oagw/src/infra/proxy/pingora_proxy.rs
```diff
@@ -0,0 +1,987 @@
...
+        let mut lb = LoadBalancer::<RoundRobin>::from_backends(backends);
+        lb.health_check_frequency = Some(Duration::from_secs(10));
+        lb.update_frequency = Some(Duration::from_secs(30));
+
+        // update() calls discover() which resolves DNS and populates both
+        // the backend selector and the addr_map in a single pass.
+        // discover() uses blocking ToSocketAddrs (same as try_from_iter),
+        // so now_or_never() is safe here.
+        use futures_util::FutureExt;
+        lb.update().now_or_never()?.ok()?;
```
MikeFalcon77: **[P2 / fragility]** `now_or_never()` succeeds only because `discover()` currently calls blocking `ToSocketAddrs`. If the DNS fix above (or any future change) makes `discover()` truly async, `now_or_never()` returns `None`, `build_entry()` returns `None`, and `select()` silently returns `None` — all backends appear dead, every request gets a 502.

No test catches this: all tests use `127.0.0.1` which resolves instantly.

**Fix:** make `build_entry` async and `.await` the update:
```rust
async fn build_entry(&self, endpoints: &[Endpoint]) -> Option<LbEntry> {
    // ...
    lb.update().await.ok()?;
    // ...
}
```

modules/system/oagw/oagw/src/infra/proxy/pingora_proxy.rs
```diff
@@ -0,0 +1,987 @@
...
+
+        // Slow path: build and cache.
+        let entry = self.build_entry(endpoints)?;
+        let backend = entry.lb.select(b"", 256)?;
+        let addr_key = backend.addr.to_string();
+        let result = {
+            let map = entry.addr_map.load();
+            map.get(&addr_key).cloned()
+        };
+        self.cache.insert(upstream_id, entry);
```
MikeFalcon77: **[P2 / race]** `select()` has a TOCTOU between the cache-miss check (line 255) and `cache.insert()` here. Two concurrent calls for the same `upstream_id` both miss the cache, both build an `LbEntry` (spawning a background task each), and the second insert silently drops the first entry — the first caller's `select()` result was computed from an `addr_map` that is now orphaned.

Exploit scenario: under burst traffic when an upstream is first accessed, N requests arrive simultaneously → N background health-check tasks are spawned, N-1 are immediately orphaned (leaked until they notice shutdown).

**Fix:** Use `DashMap::entry()` API:
```rust
use dashmap::mapref::entry::Entry;
match self.cache.entry(upstream_id) {
    Entry::Occupied(e) => { /* use e.get().lb */ }
    Entry::Vacant(e) => {
        let entry = self.build_entry(endpoints)?;
        let result = /* select from entry */;
        e.insert(entry);
        result
    }
}
```

modules/system/oagw/oagw/src/infra/proxy/pingora_proxy.rs
```diff
@@ -0,0 +1,987 @@
...
+    fn error_while_proxy(
+        &self,
+        _peer: &HttpPeer,
+        _session: &mut Session,
+        mut e: Box<pingora_core::Error>,
+        _ctx: &mut Self::CTX,
+        client_reused: bool,
+    ) -> Box<pingora_core::Error> {
+        if client_reused {
+            e.retry.decide_reuse(true);
```
MikeFalcon77: **[P2 / correctness]** `decide_reuse(true)` tells Pingora to reconnect and **resend** the request. The comment says "the request has not yet been sent", but that is not guaranteed — Pingora could have partially sent the request headers/body before the connection was reset by the peer.

For non-idempotent methods (`POST`, `DELETE`), retry-on-reuse can cause **duplicate side effects** at the upstream (e.g. double charge, double delete).

**Fix:** Guard on method idempotency:
```rust
fn error_while_proxy(
    &self,
    _peer: &HttpPeer,
    session: &mut Session,
    mut e: Box<pingora_core::Error>,
    _ctx: &mut Self::CTX,
    client_reused: bool,
) -> Box<pingora_core::Error> {
    if client_reused {
        let method = session.req_header().method.clone();
        let idempotent = matches!(
            method,
            http::Method::GET | http::Method::HEAD
            | http::Method::PUT | http::Method::DELETE
            | http::Method::OPTIONS
        );
        e.retry.decide_reuse(idempotent);
    }
    e
}
```
**Verify:** Integration test — POST through a backend that closes after keep-alive idle; assert the upstream receives exactly 1 request.

modules/system/oagw/oagw/src/infra/proxy/session_bridge.rs
```diff
@@ -0,0 +1,664 @@
...
+/// carried in `headers` are dropped to prevent duplicate framing headers.
+pub(crate) fn serialize_request_wire(
+    method: &Method,
+    url: &str,
+    headers: &HeaderMap,
+    body: Option<&Bytes>,
+) -> Vec<u8> {
+    let body_len = body.map_or(0, |b| b.len());
+    let mut buf = Vec::with_capacity(512 + body_len);
+    let _ = write!(buf, "{} {} HTTP/1.1\r\n", method, url_path_and_query(url));
```
MikeFalcon77: **[P1 / injection]** `url_path_and_query(url)` output is interpolated directly into the HTTP request line with no CRLF sanitization. If a crafted URL contains `\r\n`, an attacker who controls the upstream URL (e.g. via query params that leak into the URL) can inject arbitrary headers into the request sent to the upstream.

Exploit:
```
url = "https://victim.com/path?x=1\r\nEvil-Header: pwned\r\n"
// serialized as:
// GET /path?x=1\r\nEvil-Header: pwned\r\n HTTP/1.1\r\n
```

**Fix:** Assert or strip `\r` and `\n` from the path+query:
```rust
let pq = url_path_and_query(url);
debug_assert!(!pq.contains('\r') && !pq.contains('\n'),
    "CRLF in request URI: {pq}");
let pq = pq.replace(['\r', '\n'], "");
```

**Verify:** Unit test: `serialize_request_wire` with `\r\n` in the URL; assert output contains no bare `\r\n` except the proper line terminator.

MikeFalcon77 (commented):
## Review: oagw pingora-based proxy

Solid migration from reqwest to Pingora with a clean DuplexStream bridge, DNS-aware discovery, and well-tested endpoint selection. Several issues below, ordered by severity.

modules/system/oagw/oagw/src/infra/proxy/session_bridge.rs
mattgarmon: The exploit scenario is not reachable in the current call chain — the URL is assembled from http::Uri::path() (rejects CRLF at parse time) and form_urlencoded::Serializer (percent-encodes control characters). The endpoint host and route path are admin-configured values from the DB, not request-controlled.

Added defense-in-depth anyway: serialize_request_wire now strips CR/LF from the path+query before interpolation and logs a warning if it ever triggers. Zero allocation on the normal path. Regression test (serialize_request_strips_crlf_from_url) verifies the injected payload cannot produce a separate header line

modules/system/oagw/oagw/src/domain/services/management.rs
```diff
@@ -28,6 +32,68 @@ impl ControlPlaneServiceImpl {
...
+        {
+            return Err(DomainError::validation(format!(
+                "endpoint[{i}] uses IPv6 address '{}'; IPv6 endpoints are not yet supported",
+                ep.host
+            )));
+        }
+    }
+
+    // Check all-IP vs all-hostname consistency.
+    let ip_count = endpoints
```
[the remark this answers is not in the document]
mattgarmon: Acknowledged — the IPv4 SSRF surface is real

A hardcoded deny-list would break legitimate internal-service deployments (e.g., backends on 10.x.x.x, localhost in dev). The right fix is a configurable egress policy — e.g., deny_private_endpoints: bool in gateway config — that also covers DNS rebinding (hostname resolving to private IP at DnsDiscovery::resolve() time)

Let's address this in a separate issue.

modules/system/oagw/oagw/src/infra/proxy/service.rs
```diff
@@ -74,9 +87,98 @@ impl DataPlaneServiceImpl {
...
+            });
+        }
+
+        // Tier 1: Explicit selection via X-OAGW-Target-Host header.
+        if let Some(target_host) = req_headers
+            .get("x-oagw-target-host")
+            .and_then(|v| v.to_str().ok())
+        {
+            // Validate format: must be a hostname or IP, no port/path/special chars.
+            if target_host.is_empty()
```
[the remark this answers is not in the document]
mattgarmon: Good catch. Replaced the denylist with an allowlist — the target host must now match [a-zA-Z0-9._-] only

Tests updated to include new cases.

PR #868, opened 2026-03-04
Set `AllowInsecureHttp` by default in modkit-http config

### LINKED ISSUE #861: modkit-auth: Allow HTTP OAuth2 endpoints for internal service communication
## Summary
Right now, modkit-auth’s OAuth2 token implementation requires HTTPS for OIDC discovery endpoints. This is enforced even if the allow-insecure-http feature flag is turned on in modkit-http.

Because of that, the framework can’t be used in setups where internal services talk over plain HTTP. That’s a common pattern in Kubernetes service meshes, on-prem deployments, and other trusted internal networks.

## Error
```
OIDC discovery invalid scheme 'http': HTTPS required (transport security is TlsOnly)
```

## Root Cause
In https://github.com/cyberfabric/cyberfabric-core/blob/main/libs/modkit-auth/src/oauth2/token.rs#L69-L73, the HTTP Client is build without calling `.allow_insecure_http()`
```
let client = HttpClientBuilder::new()
    .with_timeout(timeout)
    .build()?;
```
While https://github.com/cyberfabric/cyberfabric-core/blob/main/libs/modkit-http/src/builder.rs#L122-L131 provides the `allow_insecure_http()` method, it is conditionally compiled behind `[cfg(any(debug_assertions, feature = "allow-insecure-http"))]` and is never called by `modkit-auth`, even if this feature is enabled. Credits: @Bechma 

## Expected Behaviour
Users should be able to opt-in (or opt-out, depending on chosen solution) to allow HTTP OAuth2 endpoints for trusted internal network scenarios while keeping HTTPS as the recommended default for external communication.

## Proposed Solution 
Reverse the default (opt-out)
Change the API to .disable_insecure_http() instead of .allow_insecure_http(), making HTTP the default with explicit opt-out for strict HTTPS enforcement. Credits: @MikeFalcon77

genericaccount-de:
Set `AllowInsecureHttp` by default in `modkit-http` config, add `deny_insecure_http` method for TLS enforcement.

Dev task: https://github.com/cyberfabric/cyberfabric-core/issues/861 

## Summary by CodeRabbit

* **Refactor**
  * Added an explicit TLS-enforcement option for the HTTP client; legacy per-call insecure toggle removed.
  * Client configuration defaults now allow insecure HTTP by default; TLS must be explicitly enforced.

* **Chores**
  * Build-time feature exposing insecure-HTTP was removed.

* **Tests**
  * Test setups updated to match the new default transport behavior and TLS-enforcement option.

* **Documentation**
  * Client docs and test guidance updated to reflect the new transport defaults and usage.

### DIFF
--- a/libs/modkit-auth/docs/modkit-auth-client-impl.md
+++ b/libs/modkit-auth/docs/modkit-auth-client-impl.md
@@ -18,7 +18,7 @@ token source and establishes patterns that must be followed:
 | Background refresh | `run_jwks_refresh_task` + `CancellationToken` | `aliri_tokens` watcher |
 | Backoff | Manual exponential in `calculate_backoff()` | Delegated to `aliri_tokens` |
 | Error mapping | `map_http_error(HttpError) -> ClaimsError` | Extract shared `map_http_error` or create parallel `map_http_error(HttpError) -> TokenError` |
-| Test scaffold | `httpmock` + `allow_insecure_http()` | Same pattern |
+| Test scaffold | `httpmock` | Same pattern |
 
 Key reuse points:
 
@@ -34,8 +34,7 @@ Key reuse points:
    `HttpClientConfig::token_endpoint()` instead (retries are appropriate for
    token acquisition).
 4. **Test pattern** — `jwks.rs:482-499` (`test_provider_with_http`) shows
-   how to build a test `HttpClient` with `allow_insecure_http()` for
-   `httpmock`. Reuse the same approach.
+   how to build a test `HttpClient` for `httpmock`. Reuse the same approach.
 5. **Dependencies already available** — `modkit-auth/Cargo.toml` already has
    `arc-swap`, `modkit-http`, `tokio-util`, `httpmock` (dev). Only
    `aliri_tokens` needs to be added.
@@ -361,7 +360,7 @@ Check every item:
 
 8. COMPILATION: `cargo check -p modkit-auth` succeeds.
 
-9. TESTS (use httpmock + allow_insecure_http, same pattern as
+9. TESTS (use httpmock, same pattern as
    jwks.rs test_provider_with_http at line 482):
    - Mock server returns valid token response → assert access_token + lifetime.
    - Mock returns response without expires_in → assert default_ttl used.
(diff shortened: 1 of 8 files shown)


MikeFalcon77 (commented):
## Code Review — PR #868

### Issues

#### 1. 🔴 `build()` warning is now too noisy (Medium)

`builder.rs` lines 244–250 emit a `tracing::warn!` every time `AllowInsecureHttp` is the effective transport:

```rust
if self.config.transport == TransportSecurity::AllowInsecureHttp {
    tracing::warn!(
        "insecure HTTP enabled (TransportSecurity::AllowInsecureHttp); \
         use only for testing with mock servers"
    );
}
```

Since `AllowInsecureHttp` is now the **default**, every `HttpClientBuilder::new().build()` will fire this warning. This creates significant log noise in any service that doesn't explicitly call `.deny_insecure_http()`. Moreover, the warning text ("use only for testing with mock servers") directly contradicts the design intent of this PR.

**Suggestion**: Either remove the warning entirely.

---

#### 2. 🟡 `for_testing()` doc is stale (Low)

`HttpClientConfig::for_testing()` (config.rs ~line 674) advertises "allows insecure HTTP" as its distinguishing feature:

```rust
/// Create configuration for testing with mock servers (allows insecure HTTP)
///
/// **WARNING**: This configuration allows plain HTTP connections.
/// Use only for local testing with mock servers, never in production.
```

After this PR, **all** presets allow insecure HTTP, so this doc is misleading. The real value of `for_testing()` is its relaxed timeouts/retry/pool settings — the doc should reflect that.

---

#### 3. 🟡 `modkit-auth` docs reference removed API (Low)

`libs/modkit-auth/docs/modkit-auth-client-impl.md` still references `allow_insecure_http()` in 4 places (lines 21, 37, 364, 787). While this is a design doc (not public API docs), it will confuse contributors looking at it after this PR.

---

#### 4. 🟡 Consider keeping `TlsOnly` for `token_endpoint()`

After this PR, `token_endpoint()` defaults to `AllowInsecureHttp`. OAuth2 token endpoints transmit client credentials/secrets, so they are more security-sensitive than generic HTTP calls. Consider whether `token_endpoint()` should retain `TlsOnly` as an explicit, safer default — callers can always override with `.transport(TransportSecurity::AllowInsecureHttp)` when needed.

(CodeRabbit also flagged this — agree it's worth a deliberate decision.)

PR #876, opened 2026-03-04
chore: replace parking_lot with std::sync primitives

### LINKED ISSUE #874: Remove parking_lot dependency in favor of std::sync
Since Rust 1.62, std::sync::Mutex on Linux uses a futex-based implementation that matches parking_lot performance on the uncontended fast path (CAS in userspace, no syscall). The memory footprint difference (1 byte vs ~8 bytes per lock) is negligible in practice.
related PR: https://github.com/rust-lang/rust/pull/95035

parking_lot is a third-party crate with 60+ open GitHub issues, including deadlock reports. Its only remaining practical advantage - no poisoning - is a design trade-off, not a clear win: poisoning surfaces bugs where a thread panicked while holding a lock and left shared state inconsistent.

Changes:
Replace all parking_lot::Mutex / parking_lot::RwLock with std::sync equivalents
Add .expect("poisoned") to all lock call sites
Remove parking_lot from workspace and crate Cargo.toml files
Update DEPENDENCIES.md guideline

MikeFalcon77:
Since Rust 1.62 std::sync::Mutex uses a futex-based implementation on Linux with comparable uncontended performance. parking_lot adds a third-party dependency with 60+ open issues (including deadlocks) for no meaningful benefit.

Add MutexExt/RwLockExt convenience traits (modkit-utils) that provide .hold(), .hold_read(), .hold_write() methods. These handle poisoning internally and avoid clippy unwrap_used/expect_used/missing_panics_doc noise at call sites while keeping the same panic-on-poison semantics parking_lot had.

## Release Notes

* **New Features**
  * Added per-instance REST endpoint tracking capability for enhanced configuration flexibility.

* **Chores**
  * Replaced external concurrency primitives with Rust standard library equivalents for simplified dependency management.
  * Updated synchronization patterns with ergonomic extension traits for consistent lock handling.

* **Documentation**
  * Updated synchronization guidelines to reflect current best practices and recommended patterns.


Bechma (changes requested):
Having said what I said in my previous comment, we need to be smart here, our chain of dependencies shows that parking_lot is being used in:
- dashmap (this dep can be also removed, I can explain further, but I think this is not the place)
- futures-intrusive (IDK about this dep, it's part of other deps we have from futures)
- governor (I would say this crate is unmovable
- neli (IDK, like futures-intrusive)
- referencing (IDK)
- temp-env
- tokio

So looking at the last one, we are not saving anything with this change, parking_lot dependency will still be shipped. We are using tokio for everything, so we achieve nothing with this.

PR #926, opened 2026-03-06
feat(mini-chat): return attachment summaries in list messages response

aviator5:
- Replace bare attachment_ids with rich AttachmentSummary objects (kind, filename, status, optional img_thumbnail) on Message responses.
- Add batch_attachment_summaries repo method with a single JOIN query to avoid N+1.
- Introduce attachment and message_attachment DB entities and corresponding migration tables.

## Summary by CodeRabbit

* **New Features**
  * Messages now include full attachment details (filename, type, status) and support multiple attachments.
  * Image attachments show preview thumbnails with dimensions and embedded preview data.
  * Message listing returns attachments per message for richer context.

* **Database**
  * Schema updated to store attachments and message–attachment links, enabling per-message attachment retrieval.

* **Tests**
  * Added integration and repo tests covering attachment storage, linking, thumbnails, and cross-tenant behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1224,6 +1224,7 @@ dependencies = [
  "anyhow",
  "async-trait",
  "axum",
+ "base64 0.22.1",
  "bytes",
  "cf-authz-resolver-sdk",
  "cf-mini-chat-sdk",
--- a/modules/mini-chat/mini-chat/Cargo.toml
+++ b/modules/mini-chat/mini-chat/Cargo.toml
@@ -44,6 +44,9 @@ futures = { workspace = true }
 bytes = { workspace = true }
 regex = { workspace = true }
 
+# Encoding
+base64 = { workspace = true }
+
 # Serde and JSON
 serde = { workspace = true }
 serde_json = { workspace = true }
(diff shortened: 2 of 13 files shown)


aviator5:
Re: tenant consistency on the attachment join path — I don't think this is a real isolation gap.

`chat_id` already acts as a transitive tenant boundary: every `chats` row has a single `tenant_id`, and all child rows (`messages`, `attachments`, `message_attachments`) are created through `secure_insert` scoped to the same tenant. The FK chain `message_attachments.(message_id, chat_id) → messages` and `message_attachments.(attachment_id, chat_id) → attachments` ensures all joined rows belong to the same chat, and a chat belongs to exactly one tenant.

For `attachments.tenant_id` to differ from `message_attachments.tenant_id` while sharing the same `chat_id`, you'd need an attachment created under a different tenant for the same chat — which is impossible given the `AccessScope`-gated insert path.

The query already filters `message_attachments.tenant_id` (line 214), which is sufficient. Adding a redundant `attachments.tenant_id` predicate would be dead code — no rows can ever be excluded by it.

PR #1123, opened 2026-03-14
Add profiling capabilities with separate http server.

genericaccount-de:
Add profiling capabilities with separate http server.

## Summary by CodeRabbit

* **New Features**
  * Added on-demand CPU and heap profiling via HTTP endpoints when enabled in configuration.
  * Profiling server binds to configurable address (default: 127.0.0.1:6060) with optional bearer token authentication.
  * Supports multiple output formats: pprof protobuf and flamegraph visualization.

* **Documentation**
  * Added comprehensive profiling guide covering configuration, endpoints, security, and platform support.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,7 +90,10 @@ resolver = "3"
 [workspace.lints.rust]
 deprecated = "warn"
 non_ascii_idents = "forbid"
-unsafe_code = "forbid"
+# `deny` (not `forbid`) so that the Windows profiling sampler in modkit can
+# locally `#[allow(unsafe_code)]` for Win32 FFI.  Every other module remains
+# effectively forbidden because `deny` triggers the same compile error.
+unsafe_code = "deny"
 unused_mut = "warn"
 noop_method_call = "warn"
 unused_import_braces = "warn"
@@ -505,6 +508,10 @@ nanoid = "0.4"
 
 # Metrics
 tokio-metrics = { version = "0.4", features = ["rt"] }
+
+# Profiling (CPU sampling)
+pprof = { version = "0.15", default-features = false, features = ["flamegraph"] }
+backtrace = "0.3"
 
 # GTS dependencies
 jsonschema = "0.40"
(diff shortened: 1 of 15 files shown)


libs/modkit/src/bootstrap/host/profiling/server.rs
```diff
@@ -0,0 +1,429 @@
...
+         Available profiles:\n\
+         - /debug/pprof/profile?seconds=30&format=protobuf  CPU profile\n\
+         - /debug/pprof/profile?seconds=10&format=flamegraph CPU flamegraph SVG\n\
+         - /debug/pprof/heap                                 Heap / process memory stats\n"
+    );
+
+    (StatusCode::OK, [(header::CONTENT_TYPE, "text/plain")], body)
+}
+
+async fn cpu_profile_handler(
```
MikeFalcon77: **HIGH**

No concurrency guard prevents multiple simultaneous CPU profile captures.

On Unix, `pprof-rs` uses `SIGPROF` which is process-global — concurrent profiling sessions will interfere with each other and produce corrupt results. On Windows, multiple sampler threads will compete.

Add an `Arc<tokio::sync::Mutex<()>>` (or `AtomicBool` with try-acquire) in `ProfileState` and return `409 Conflict` if a profile is already in progress.

Cargo.toml
```diff
@@ -90,7 +90,10 @@ resolver = "3"
 [workspace.lints.rust]
 deprecated = "warn"
 non_ascii_idents = "forbid"
-unsafe_code = "forbid"
+# `deny` (not `forbid`) so that the Windows profiling sampler in modkit can
+# locally `#[allow(unsafe_code)]` for Win32 FFI.  Every other module remains
+# effectively forbidden because `deny` triggers the same compile error.
+unsafe_code = "deny"
```
MikeFalcon77: **HIGH**

Downgrading `unsafe_code` from `forbid` to `deny` at the **workspace** level means every crate in the workspace can now locally `#[allow(unsafe_code)]` — not just the profiling module.

The blast radius is the entire workspace. Any future code can silently opt into unsafe without a compile error.

Scope this more narrowly: keep workspace-level `forbid`, and override to `deny` only in `libs/modkit/Cargo.toml` via `[lints.rust] unsafe_code = "deny"`. That limits the opt-out to the single crate that needs it.

MikeFalcon77:
please create separate PR. Moreover I have an idea to create separate modkit-debug crate with all stuff like: pprof, tokio-console, best practices for pprof/crash dumps and etc.

PR #1213, opened 2026-03-17
fix(mini-chat): eliminate post-commit DB read in mutation stream setup

mattgarmon:
The retry/edit handler called `get_chat()` after the mutation transaction committed, solely to fetch `chat.model` for provider resolution. Under SQLite write contention from a prior turn's background finalization (quota settlement, CAS writes), this read could fail with "database is locked", returning a transient 500 to the client.
 
Fixes: transient 500 on `test_edit_emits_turn_started_with_new_request_id`

## Summary by CodeRabbit

* **Refactor**
  * Streamlined chat mutation operations to utilize payload data directly.
  * Enhanced mutation flow with improved data threading through validation and result stages.
  * Updated mutation result structure to propagate model information across delete, retry, and edit operations.

### DIFF
--- a/modules/mini-chat/mini-chat/src/domain/service/turn_service.rs
+++ b/modules/mini-chat/mini-chat/src/domain/service/turn_service.rs
@@ -102,6 +102,9 @@ pub struct MutationResult {
     /// Snapshot boundary computed before the new user message was persisted.
     /// Ensures deterministic context assembly (DESIGN `§ContextPlan` Determinism P1).
     pub snapshot_boundary: Option<crate::domain::repos::SnapshotBoundary>,
+    /// Chat model carried from the mutation transaction so the handler can
+    /// resolve the provider without a redundant DB round-trip.
+    pub chat_model: String,
 }
 
 // ════════════════════════════════════════════════════════════════════════════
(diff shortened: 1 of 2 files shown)


modules/mini-chat/mini-chat/src/domain/service/turn_service.rs
```diff
@@ -102,6 +102,9 @@ pub struct MutationResult {
     /// Snapshot boundary computed before the new user message was persisted.
     /// Ensures deterministic context assembly (DESIGN `§ContextPlan` Determinism P1).
     pub snapshot_boundary: Option<crate::domain::repos::SnapshotBoundary>,
+    /// Chat model carried from the mutation transaction so the handler can
+    /// resolve the provider without a redundant DB round-trip.
```
MikeFalcon77: **LOW**

`validate_mutation` now returns a 3-element positional tuple `(AccessScope, TurnModel, String)`. With three fields the meaning of the last element is opaque at call sites (`_chat_model` in delete, unnamed destructuring elsewhere).

Not a blocker for this PR — but if this function grows another return value, extract a named struct (e.g. `ValidatedMutation { scope, target, chat_model }`) to keep call sites self-documenting.

MikeFalcon77 (approved):

PR #1241, opened 2026-03-19
fix(minichat): Fix 500 on second attachment upload

jalankulkija:
In PostgreSQL, SUM(bigint) returns NUMERIC, not INT8. The sum_size_bytes query decoded the result into Option<i64> (mapped as INT8), which worked on the first upload — because an empty table returns NULL and sqlx skips type checking for NULL — but failed on the second upload when a real non-NULL NUMERIC value was returned.

Fixed by casting: SUM(size_bytes)::bigint.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved attachment storage size calculations by enhancing database data type handling, ensuring accurate file size reporting and better support for larger attachments.


MikeFalcon77 (approved):

MikeFalcon77:
please add PgSQL tests for mini-chat via test containers

PR #1373, opened 2026-03-30
docs(account-management) PRD + DESIGN proposal

diffora:
## Summary by CodeRabbit

* **Documentation**
  * Added multiple ADRs: IdP/AuthN contract split, IdP as source-of-truth for users, IdP user→tenant binding, read-time metadata inheritance (with barrier-aware semantics), canonical mode-conversion endpoints/approval flow, and rationale rejecting Resource Group as canonical tenant store. Several ADRs mark superseded predecessors and record confirmation criteria and consequences.

* **Chores**
  * Added a tenant-type schema exposing trait configuration: root capability, allowed parent types, and IdP provisioning flag.

### DIFF
--- a/modules/system/account-management/docs/DESIGN.md
+++ b/modules/system/account-management/docs/DESIGN.md
@@ -1,0 +1,1507 @@
+Created:  2026-04-01 by Diffora
+
+# Technical Design — Account Management (AM)
+
+- [ ] `p3` - **ID**: `cpt-cf-account-management-design-am`
+
+<!-- toc -->
+
+- [1. Architecture Overview](#1-architecture-overview)
+  - [1.1 Architectural Vision](#11-architectural-vision)
+  - [1.2 Architecture Drivers](#12-architecture-drivers)
+  - [1.3 Architecture Layers](#13-architecture-layers)
+- [2. Principles & Constraints](#2-principles--constraints)
+  - [2.1 Design Principles](#21-design-principles)
+  - [2.2 Constraints](#22-constraints)
+- [3. Technical Architecture](#3-technical-architecture)
+  - [3.1 Domain Model](#31-domain-model)
+  - [3.2 Component Model](#32-component-model)
+  - [3.3 API Contracts](#33-api-contracts)
+  - [3.4 Internal Dependencies](#34-internal-dependencies)
+  - [3.5 External Dependencies](#35-external-dependencies)
+  - [3.6 Interactions & Sequences](#36-interactions--sequences)
+  - [3.7 Database schemas & tables](#37-database-schemas--tables)
+  - [3.8 Error Codes Reference](#38-error-codes-reference)
+- [4. Additional Context](#4-additional-context)
+  - [4.1 Applicability and Delegations](#41-applicability-and-delegations)
+  - [4.2 Security Architecture](#42-security-architecture)
+  - [Threat Modeling](#threat-modeling)
+  - [4.3 Reliability and Operations](#43-reliability-and-operations)
+  - [Data Governance](#data-governance)
+  - [Testing Architecture](#testing-architecture)
+  - [Open Questions](#open-questions)
+  - [Documentation Strategy](#documentation-strategy)
+  - [Known Limitations & Technical Debt](#known-limitations--technical-debt)
+- [5. Traceability](#5-traceability)
+
+<!-- /toc -->
+
+> **Abbreviation**: Account Management = **AM**. Used throughout this document.
+
+## 1. Architecture Overview
+
+### 1.1 Architectural Vision
+
+AM is the foundational multi-tenancy source-of-truth module for the Cyber Fabric platform. It owns the tenant hierarchy, tenant type enforcement, barrier metadata, delegated IdP user operations, and extensible tenant metadata. AM follows the standard ModKit module pattern under `modules/system/account-management/`: a planned SDK crate (`account-management-sdk`) exposes transport-agnostic traits and models, and a planned implementation crate (`account-management`) provides the module lifecycle, REST API, domain logic, and infrastructure adapters.
+
+The architecture separates data ownership from enforcement. AM stores and validates the tenant tree structure, barrier flags, and type constraints. It does not evaluate authorization policies, generate SQL predicates, or validate bearer tokens on the per-request path. Tenant Resolver and AuthZ Resolver consume AM source-of-truth data for runtime enforcement. This separation keeps AM focused on administrative correctness while letting specialized resolvers optimize the hot path independently.
+
+IdP integration uses the CyberFabric gateway + plugin pattern, analogous to AuthN Resolver (see `cpt-cf-account-management-adr-idp-contract-separation`). AM defines an `IdpProviderPluginClient` trait for tenant and user administrative operations (tenant provisioning/deprovisioning, user provision/deprovision, tenant-scoped query). The plugin is discovered via GTS types-registry and resolved through `ClientHub`. The platform ships a default provider plugin; vendors substitute their own implementation behind the same trait. The IdP provider plugin is intentionally separate from the AuthN Resolver plugin — the two target different concerns (admin operations vs hot-path token validation) with different performance profiles and protocols. The contract is one-directional: AM calls IdP, IdP does not call AM.
+
+User group management is handled by the [Resource Group](../../resource-group/docs/PRD.md) module. AM registers a dedicated Resource Group type for user groups during module initialization; consumers call `ResourceGroupClient` directly for group lifecycle, membership, and hierarchy operations.
+
+#### System Context
+
+```mermaid
+graph LR
+    Admin["Platform / Tenant<br>Administrator"] -->|REST API| AM["Account<br>Management"]
+    AM -->|provision/deprovision<br>users & tenants| IdP["IdP Provider<br>Plugin"]
+    AM -->|type validation<br>& schema queries| GTS["GTS Types<br>Registry"]
+    AM -->|user-group type<br>registration & cleanup| RG["Resource<br>Group"]
+    AM ---|source-of-truth<br>tenant data| DB[("PostgreSQL")]
+    TR["Tenant Resolver"] -.->|reads tenant<br>source-of-truth| DB
+    AuthZ["AuthZ Resolver<br>(PDP plugin)"] -.->|tenant hierarchy<br>& barrier queries| TR
+    Billing["Billing System"] -.->|reads tenant<br>metadata| AM
+```
+
+**System actors by PRD ID**
+
+- `cpt-cf-account-management-actor-tenant-resolver` consumes AM source-of-truth tenant hierarchy and barrier state through the tenant data contract.
+- `cpt-cf-account-management-actor-authz-resolver` consumes tenant context, barrier inputs, and metadata authorization attributes for access decisions.
+- `cpt-cf-account-management-actor-billing` consumes read-only hierarchy and billing-relevant metadata views under platform-authorized barrier-bypass policy.
+
+### 1.2 Architecture Drivers
+
+#### Functional Drivers
+
+| Requirement | Design Response |
+|-------------|-----------------|
+| `cpt-cf-account-management-fr-root-tenant-creation` | Bootstrap runs as the first action inside the module's `lifecycle(entry = ...)` method, creating the initial root tenant with IdP linking before signalling ready. |
+| `cpt-cf-account-management-fr-root-tenant-idp-link` | Bootstrap calls `provision_tenant` for the root tenant (same contract as all tenants), forwarding deployer-configured `root_tenant_metadata` so the IdP provider plugin can establish the tenant-to-IdP binding. Any provider-returned `ProvisionResult` metadata is persisted as tenant metadata. AM does not validate binding sufficiency — binding establishment is the provider's responsibility, whether via returned metadata, external configuration, or convention. |
+| `cpt-cf-account-management-fr-bootstrap-idempotency` | Bootstrap checks for existing root tenant before creation; no-op when already present. |
+| `cpt-cf-account-management-fr-bootstrap-ordering` | Bootstrap retries IdP availability with configurable backoff and timeout before proceeding. |
+| `cpt-cf-account-management-fr-create-child-tenant` | `TenantService::create_child_tenant` validates parent status, GTS type constraints, and depth threshold. |
+| `cpt-cf-account-management-fr-hierarchy-depth-limit` | Configurable advisory threshold with optional strict mode; depth computed from parent at creation time. |
+| `cpt-cf-account-management-fr-tenant-status-change` | `TenantService::update_status` applies `active` ↔ `suspended` transitions without cascading to children. Transition to `deleted` is rejected — deletion goes through `TenantService::delete_tenant` which enforces child/resource-ownership preconditions. |
+| `cpt-cf-account-management-fr-tenant-soft-delete` | `TenantService::delete_tenant` validates the target tenant is non-root, has no non-deleted children, and has no remaining tenant-owned resource associations in the Resource Group ownership graph before soft delete; schedules hard deletion after retention period. |
+| `cpt-cf-account-management-fr-children-query` | Paginated direct-children query with status filtering via OData `$filter` on `parent_id` index. |
+| `cpt-cf-account-management-fr-tenant-read` | `TenantService::get_tenant` returns tenant details by identifier within the caller's authorized scope. |
+| `cpt-cf-account-management-fr-tenant-update` | `TenantService::update_tenant` mutates only `name` and `status` (`active` ↔ `suspended`); immutable hierarchy-defining fields are rejected; `status=deleted` is rejected with `422` — use `DELETE` endpoint. |
+| `cpt-cf-account-management-fr-tenant-type-enforcement` | `TenantService` queries `TypesRegistryClient` for type constraints at child creation time. |
+| `cpt-cf-account-management-fr-tenant-type-nesting` | Same-type nesting permitted when GTS type definition allows it; acyclicity guaranteed by tree structure. |
+| `cpt-cf-account-management-fr-managed-tenant-creation` | Tenant created with `self_managed=false`; no barrier flag set. |
+| `cpt-cf-account-management-fr-self-managed-tenant-creation` | Tenant created with `self_managed=true`; barrier flag stored for downstream resolver consumption. |
+| `cpt-cf-account-management-fr-mode-conversion-approval` | `ConversionService` owns the dual-consent lifecycle for any post-creation toggle of `tenants.self_managed`. Each side acts from its own authorized scope and root tenants are excluded from the flow. |
+| `cpt-cf-account-management-fr-mode-conversion-expiry` | A background expiry task closes unresolved conversion requests after the configured approval window without changing tenant mode. |
+| `cpt-cf-account-management-fr-mode-conversion-single-pending` | A partial unique invariant on the conversion store plus service-level conflict handling ensure at most one pending conversion request per tenant. |
+| `cpt-cf-account-management-fr-mode-conversion-consistent-apply` | Approval updates both conversion status and tenant barrier state as one consistent transaction outcome. |
+| `cpt-cf-account-management-fr-conversion-creation-time-self-managed` | `TenantService::create_tenant` accepts `self_managed=true` directly at creation time and stores the flag without a `ConversionRequest`; the parent's explicit creation call is the consent. Only post-creation toggles are routed through `ConversionService`. |
+| `cpt-cf-account-management-fr-child-conversions-query` | `ConversionService::list_inbound_for_parent` joins `conversion_requests` with `tenants` on `parent_id`. Operates within parent tenant AuthZ scope; no barrier bypass required. Exposes only conversion-request metadata (child `id`, child `name`, `initiator_side`, `target_mode`, `status`, timestamps), not full child tenant data. |
+| `cpt-cf-account-management-fr-conversion-cancel` | `ConversionService::cancel` transitions a pending `ConversionRequest` to `cancelled` only when `caller_side == initiator_side`. Exposed via `PATCH .../conversions/{r}` (child scope) and `PATCH .../child-conversions/{r}` (parent scope) with body `{"status": "cancelled"}`. Role-check failures return `409 conflict` with sub-code `invalid_actor_for_transition`. |
+| `cpt-cf-account-management-fr-conversion-reject` | `ConversionService::reject` transitions a pending `ConversionRequest` to `rejected` only when `caller_side != initiator_side`. Exposed via the same `PATCH` endpoints with body `{"status": "rejected"}`. Role-check failures return `409 conflict` with sub-code `invalid_actor_for_transition`. |
+| `cpt-cf-account-management-fr-conversion-retention` | Background job `ConversionService::soft_delete_resolved` stamps `deleted_at` on resolved rows older than `resolved_retention` (default 30d); default queries filter `deleted_at IS NULL`. Hard-delete follows AM's existing retention cadence. |
+| `cpt-cf-account-management-fr-idp-tenant-provision` | Tenant creation uses a saga pattern: (1) short TX inserts the tenant with `status=provisioning`, (2) `IdpProviderPluginClient::provision_tenant` is called outside any transaction, (3) a second short TX persists provider-returned metadata and transitions the tenant to `active`. If the IdP call fails, a compensating TX deletes the `provisioning` row — no orphan in either system. If the finalization TX fails, AM does not retry the DB completion step; the tenant remains in `provisioning` state until the background reaper compensates (see Reliability Architecture — Data Consistency). `POST /tenants` remains intentionally non-idempotent. |
+| `cpt-cf-account-management-fr-idp-tenant-provision-failure` | The tenant-creation saga distinguishes clean compensation from ambiguous external outcomes and maps both paths into deterministic public failure behavior plus reaper-backed reconciliation. |
+| `cpt-cf-account-management-fr-idp-tenant-deprovision` | Background hard-deletion job calls `IdpProviderPluginClient::deprovision_tenant` for every hard-deleted tenant. Provider implementations clean up tenant-scoped IdP resources, guided by tenant type traits such as `idp_provisioning`. Providers **MUST NOT** silently no-op on mutating operations — unsupported deprovisioning **MUST** fail with `idp_unsupported_operation`. Failure retries rather than skips. |
+| `cpt-cf-account-management-fr-idp-user-provision` | `IdpProviderPluginClient::create_user` with tenant scope binding and resolved tenant metadata (for IdP context resolution, e.g., effective Keycloak realm). |
+| `cpt-cf-account-management-fr-idp-user-deprovision` | `IdpProviderPluginClient::delete_user` with session revocation; an already-absent IdP user is treated as a successful no-op so `DELETE /tenants/{id}/users/{user_id}` remains idempotent. |
+| `cpt-cf-account-management-fr-idp-user-query` | `IdpProviderPluginClient::list_users` with tenant filter; supports optional user-ID filter for single-user lookups. |
+| `cpt-cf-account-management-fr-user-group-rg-type` | `AccountManagementModule` idempotently registers the user-group Resource Group type `gts.x.core.rg.type.v1~x.core.am.user_group.v1~` during module initialization, with `allowed_memberships` including the platform user resource type (`gts.x.core.am.user.v1~`). |
+| `cpt-cf-account-management-fr-user-group-lifecycle` | Consumers call `ResourceGroupClient` directly for group create/update/delete. AM does not proxy these operations. |
+| `cpt-cf-account-management-fr-user-group-membership` | Consumers call `ResourceGroupClient` directly for membership add/remove. Callers verify user existence via AM's user-list endpoint; RG treats `resource_id` as opaque. |
+| `cpt-cf-account-management-fr-nested-user-groups` | Nested groups via Resource Group parent-child hierarchy; cycle detection enforced by Resource Group forest invariants. No AM involvement at runtime. |
+| `cpt-cf-account-management-fr-tenant-metadata-schema` | `MetadataService` validates metadata payloads against the GTS-registered schema identified by `schema_id`, using the schema's `inheritance_policy` trait to drive resolution. |
+| `cpt-cf-account-management-fr-tenant-metadata-crud` | `MetadataService` provides CRUD for metadata entries keyed by `(tenant_id, schema_id)` with GTS schema validation. |
+| `cpt-cf-account-management-fr-tenant-metadata-api` | `MetadataService::resolve` walks the hierarchy when the schema's `inheritance_policy` trait is `inherit`; returns the tenant's own value (or empty) when `override_only`. |
+| `cpt-cf-account-management-fr-tenant-metadata-list` | `MetadataService::list_for_tenant` returns paginated own-entries for a tenant; REST endpoint `GET /api/account-management/v1/tenants/{id}/metadata` is tenant-scope-filtered by the platform layer, so self-managed barriers apply without AM-specific logic. |
+| `cpt-cf-account-management-fr-tenant-metadata-permissions` | REST handlers pass `schema_id` into `PolicyEnforcer::enforce` as a resource attribute (`SCHEMA_ID`) on `Metadata.read`, `Metadata.write`, `Metadata.delete`, and `Metadata.list` actions, so external AuthZ policy can express per-`schema_id` grants without AM evaluating policy itself. |
+| `cpt-cf-account-management-fr-deterministic-errors` | Unified error mapper translates domain and infrastructure failures to stable public categories; the authoritative HTTP/sub-code mapping is published in the OpenAPI contract. |
+| `cpt-cf-account-management-fr-observability-metrics` | OpenTelemetry metrics for domain-internal latencies (IdP calls, GTS validation, metadata resolution, bootstrap), background job throughput, error rates, and security counters. Per-endpoint CRUD counts and children-query latency are captured by platform HTTP middleware; capacity gauges (active tenants) are derivable from DB queries. Projection freshness is a Tenant Resolver concern. |
+
+#### NFR Allocation
+
+| NFR ID | NFR Summary | Allocated To | Design Response | Verification Approach |
+|--------|-------------|--------------|-----------------|----------------------|
+| `cpt-cf-account-management-nfr-context-validation-latency` | End-to-end tenant-context validation p95 ≤ 5ms | Schema design + indexes on `tenants` table (AM contribution); Tenant Resolver caching (resolver contribution) | Composite indexes on `(parent_id, status)`, `(id, status)`, and `(tenant_type_uuid)`; denormalized `depth` column avoids recursive queries. AM provides the indexed source-of-truth schema; the end-to-end p95 ≤ 5ms target requires Tenant Resolver's caching layer on top. | AM: integration tests verify indexed query baseline with seeded dataset. Platform: pre-GA load test benchmark (end-to-end through Tenant Resolver + caching) against approved deployment profile. |
+| `cpt-cf-account-management-nfr-tenant-isolation` | Zero cross-tenant data leaks | SecureConn + PolicyEnforcer on all API endpoints | All database access through `SecureConn` with tenant-scoped queries; `PolicyEnforcer` PEP pattern on every REST handler. | Automated security test suite with cross-tenant access attempts |
+| `cpt-cf-account-management-nfr-audit-completeness` | 100% tenant config changes audited | Platform append-only audit infrastructure | AM relies on the platform append-only audit infrastructure as the single audit sink. Request handlers emit audit records via the platform request pipeline; AM-owned non-request flows emit into the same sink with `actor=system` (bootstrap completion, conversion expiry, provisioning-reaper compensation, hard-delete / tenant-deprovision cleanup). Database-level `created_at`/`updated_at` timestamps provide additional chronology but are not the audit system. | Verify platform audit entries exist for state-changing API operations and for AM-owned system-actor lifecycle events in integration tests |
+| `cpt-cf-account-management-nfr-barrier-enforcement` | Barrier state sufficient for downstream enforcement; AM-owned barrier-state changes audited | `self_managed` column + tenant hierarchy in `tenants` table | AM exposes `self_managed` flag and parent-child relationships; Tenant Resolver and AuthZ Resolver consume this data for barrier traversal and access decisions. AM commits the flag synchronously within the conversion transaction; enforcement latency depends on how quickly Tenant Resolver's projection reflects the change (sync interval is a Tenant Resolver concern, not an AM guarantee). Platform request audit logging captures all barrier-state-changing operations (mode conversions) per `cpt-cf-account-management-nfr-audit-completeness`; cross-tenant access auditing is a platform AuthZ concern. | Integration tests validating barrier data completeness for resolver consumption; verify platform audit log entries exist for mode conversion operations |
+| `cpt-cf-account-management-nfr-tenant-model-versatility` | Both managed and self-managed in same tree | `self_managed` boolean per tenant, independent of siblings | Sibling tenants under the same parent can have different `self_managed` values; mode conversion is a per-tenant operation. | Integration tests with mixed-mode hierarchies |
+| `cpt-cf-account-management-nfr-compatibility` | No breaking changes within minor release | Path-based API versioning + stable SDK trait contract | REST API uses `/api/account-management/v1/` prefix; SDK trait changes require new major version with migration path. | Contract tests on SDK trait + API schema regression tests |
+| `cpt-cf-account-management-nfr-production-scale` | Approved deployment profile before DESIGN sign-off | Schema design + index strategy | Approved deployment profile: 100K tenants, depth 5 (advisory threshold 10), 300K users (IdP-stored), 30K user groups / 300K memberships (RG-stored), 1K rps peak. All targets within planning envelope. Schema impact assessment confirms existing indexes and B-tree depths are sufficient for that profile; no partitioning is required. | Capacity test against approved profile (100K tenants, 300K users, 1K rps) |
+| `cpt-cf-account-management-nfr-data-classification` | Persisted data classes documented; no credentials/profile PII stored by AM | Security architecture + data model boundaries | Tenant hierarchy metadata is classified as commercially sensitive; metadata schema classification is derived per `schema_id`; AM persists IdP-issued UUID identity references only where required for traceability and never stores credentials or IdP profile data outside the platform audit infrastructure. | Design review of persisted fields and audit payloads; integration checks confirm user lifecycle paths do not create local user/profile records |
+| `cpt-cf-account-management-nfr-reliability` | Reads stay available during IdP outages; retry contract is explicit | Saga-based tenant creation + degraded-mode reads + reaper compensation | Non-IdP-dependent reads and admin operations continue during IdP outages. Tenant creation uses the three-step provisioning saga; clean step-2 compensation returns `idp_unavailable` and is retryable, while transport failure, timeout, or finalization failure after IdP-side success leave `POST /tenants` in an ambiguous, non-idempotent state that callers must reconcile before retry. | Integration tests for clean compensation, ambiguous finalization failure, provisioning reaper cleanup, and degraded reads during IdP outage |
+| `cpt-cf-account-management-nfr-data-lifecycle` | Tenant deprovisioning cascades cleanup | `TenantService::delete_tenant` + background hard-delete job | Soft delete transitions to `deleted` status; background job hard-deletes after retention period; cascade triggers IdP `deprovision_tenant` (with retry on failure), Resource Group cleanup for tenant-scoped user groups (via `ResourceGroupClient`), and metadata entry deletion. | Integration tests verifying cascaded cleanup |
+| `cpt-cf-account-management-nfr-authentication-context` | Authenticated requests via platform SecurityContext; MFA for admin ops deferred to platform AuthN policy | `SecurityContext` requirement on all REST handlers via framework middleware | AM does not validate tokens or enforce MFA directly. All REST endpoints require a valid `SecurityContext` provided by the framework AuthN pipeline. MFA enforcement for administrative operations such as tenant creation and mode conversion is a platform AuthN policy concern — AM relies on the framework to reject requests that do not meet the configured authentication strength. | API tests: every endpoint returns 401 without valid `SecurityContext`; E2E: admin operations succeed only with authenticated requests |
+| `cpt-cf-account-management-nfr-data-quality` | Transactional commit visibility for Tenant Resolver sync; hierarchy integrity checks | Transactional DB writes + diagnostic integrity query | AM commits hierarchy changes transactionally, making them immediately visible in the `tenants` table for Tenant Resolver consumption. Schema stability per `cpt-cf-account-management-nfr-compatibility` ensures the database-level data contract remains intact. AM provides a hierarchy integrity check (orphaned children, broken parent references, depth mismatches) as a diagnostic capability. The end-to-end 30s freshness SLO is a platform-level target requiring Tenant Resolver's sync mechanism; AM's contribution is transactional commit visibility. | Integration: hierarchy integrity check detects seeded anomalies (orphaned child, depth mismatch); Integration: committed writes are immediately visible via direct table query |
+| `cpt-cf-account-management-nfr-data-integrity-diagnostics` | Diagnostic checks for observable hierarchy anomalies | `TenantService::check_hierarchy_integrity()` + observability surface | AM exposes explicit integrity diagnostics for orphaned children, broken parent links, root-count anomalies, and depth mismatches. | Integration tests seed anomalies and verify diagnostic output plus metric surfacing |
+| `cpt-cf-account-management-nfr-data-remediation` | Operator-visible remediation path for AM-owned integrity anomalies | Observability + runbook-owned lifecycle handling | Compensation failures and integrity anomalies emit telemetry quickly, remain visible until addressed, and map to runbook-driven triage owned by platform operations. | Alert simulation and operational review |
+| `cpt-cf-account-management-nfr-ops-metrics-treatment` | Minimum operational treatment for AM domain metrics | Shared dashboards + alert routing | AM publishes the minimum metric set required for operator treatment: IdP failures, bootstrap not-ready, provisioning reaper activity, integrity violations, and cleanup failures. | Dashboard/alert review plus smoke checks in staging |
+
+#### Key ADRs
+
+The following architecture decisions are adopted in this DESIGN:
+
+| Decision Area | Adopted Approach | ADR |
+|---------------|-----------------|-----|
+| IdP contract design | Separate IdP provider plugin (`IdpProviderPluginClient`) from AuthN Resolver plugin, both following CyberFabric gateway + plugin pattern with independent GTS schemas. | `cpt-cf-account-management-adr-idp-contract-separation` — [ADR-0001](ADR/0001-cpt-cf-account-management-adr-idp-contract-separation.md) |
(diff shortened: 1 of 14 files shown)


modules/system/accounts/docs/PRD.md
```diff
@@ -0,0 +1,1560 @@
...
+
+The IdP contract is one-directional: HAS calls IdP. IdP does not call HAS. HAS tolerates IdP unavailability during bootstrap with retry/backoff.
+
+The IdP integration contract is intentionally separate from the AuthN Resolver contract. See [ADR-0001: IdP Contract Separation](./ADR/0001-idp-contract-separation.md) for rationale.
+
+### 3.3 Barrier Tenant Isolation
+
+#### Responsibility Split: HAS stores, Tenant Resolver + AuthZ enforce
+
+**HAS treats barrier purely as data.** For tenants that are self-managed, HAS stores and returns the self-managed flag in API responses; nothing more. All HAS queries return data regardless of barrier values — barrier enforcement is entirely outside HAS scope.
```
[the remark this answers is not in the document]
diffora: Yes, tenant barriers must be respected for user management operations. 
And this is enforced automatically by the standard platform AuthZ path — all HAS REST API endpoints (including those that trigger IdP user operations) pass through PolicyEnforcer → AuthZ Resolver → Tenant Resolver, which applies barrier logic during tenant hierarchy traversal.

modules/system/accounts/docs/ADR/0001-idp-contract-separation.md
```diff
@@ -0,0 +1,87 @@
...
+* Good, because a single contract simplifies the integration surface — one implementation per IdP.
+* Good, because shared state (connection pools, configuration) is implicit.
+* Bad, because hot-path token validation becomes coupled to transactional admin logic, risking latency degradation on every API request.
+* Bad, because changes to admin protocols force changes to the authentication contract.
+* Bad, because deployments that don't need user management still carry the admin contract surface.
+
+## Traceability
+
+- **PRD**: [PRD.md](../PRD.md)
+- **DESIGN**: [DESIGN.md](../DESIGN.md)
```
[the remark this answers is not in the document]
diffora: Removed link.

modules/system/accounts/docs/PRD.md
```diff
@@ -0,0 +1,1568 @@
...
+
+### 1.2 Background / Problem Statement
+
+Cyber Fabric needs a unified multi-tenancy model that supports diverse organizational structures and business models within a single deployment. Without a shared tenant hierarchy, each organizational model requires separate infrastructure or custom integration, increasing operational cost and limiting platform scalability.
+
+The platform needs a tenant model with unlimited hierarchy depth (configurable advisory threshold), visibility barriers for self-managed tenants, and consistent tenant context propagation across all services.
+
+### 1.3 Goals (Business Outcomes)
+
+- Provide one stable contract for tenant hierarchy, type enforcement, barrier semantics, and user group coordination (via RG).
```
ast2074: This is a PRD - then it should provide a list of product / business scenarios.
W/o that it's hard to deside / prove if the tenant hierarchy is ever required, for example...

modules/system/accounts/docs/PRD.md
```diff
@@ -0,0 +1,1568 @@
...
+
+**Success criteria:**
+
+| Metric | Baseline | Target | Timeframe |
+|--------|----------|--------|-----------|
+| Tenant onboarding effort | Manual multi-step provisioning per organizational model | Single API call — tenant create with type and mode | Module GA |
+| Separate deployments per tenant model | One deployment per managed/self-managed pattern | Both modes in one tenant tree — zero per-model deployments | Module GA |
+| End-to-end tenant context validation latency (p95) | N/A (no unified solution) | ≤ 5ms with caching in the resolver path | Pre-GA load test gate |
+| Cross-tenant data leaks | N/A | Zero — verified by automated security test suite | Pre-GA security gate |
+| Hierarchy depth coverage | Fixed shallow hierarchies (2–3 levels) | Unlimited depth with configurable advisory threshold (default: 10); validated with at least 3 distinct type topologies including hierarchies beyond the default threshold | Pre-GA integration test gate |
```
ast2074: Unlimited hierarchy will hurt some performance optimization in dependent domain implementations, esp. in case of large data volumes, so this requirement should better be grounded in advance by corresponding perf benches...

modules/system/accounts/docs/PRD.md
```diff
@@ -0,0 +1,1568 @@
...
+- Enable end-to-end tenant-context validation to achieve p95 ≤ 5ms on every API call.
+- Maintain zero cross-tenant data leaks, verified by automated security tests.
+- Provide a pluggable IdP integration contract for user lifecycle operations.
+
+**Success criteria:**
+
+| Metric | Baseline | Target | Timeframe |
+|--------|----------|--------|-----------|
+| Tenant onboarding effort | Manual multi-step provisioning per organizational model | Single API call — tenant create with type and mode | Module GA |
+| Separate deployments per tenant model | One deployment per managed/self-managed pattern | Both modes in one tenant tree — zero per-model deployments | Module GA |
```
ast2074: What does 'deployment' mean here?

modules/system/accounts/docs/PRD.md
```diff
@@ -0,0 +1,1568 @@
...
+| **MSP / Managed Services** | A managed service provider operates customer environments on their behalf. | Parent-child managed relationship (no barrier). Parent admin impersonates child tenant for support operations with time-bounded sessions and full audit trail. |
+| **Distributor / Reseller** | A distributor resells platform capacity but does not access customer data. | Self-managed child tenants create visibility barriers. Distributor sees billing metadata (`BarrierMode::Ignore`) but cannot access customer APIs or resources. |
+
+**IdP integration convention.** HAS defines an abstract IdP contract for user lifecycle operations (provision, deprovision). The contract is pluggable — deployments can use any conforming IdP implementation (Keycloak, Azure AD, Okta, custom). HAS owns the tenant model; IdP owns user authentication and token issuance. HAS calls IdP via the pluggable contract for user-tenant binding operations.
```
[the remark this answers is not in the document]
diffora: Users and tenants are mapped on IDP.
see: https://github.com/diffora/cyberfabric-core/blob/main/modules/system/account-management/docs/ADR/0008-cpt-cf-accounts-adr-idp-user-tenant-binding.md

modules/system/accounts/docs/PRD.md
```diff
@@ -0,0 +1,1568 @@
...
+
+Cyber Fabric needs a unified multi-tenancy model that supports diverse organizational structures and business models within a single deployment. Without a shared tenant hierarchy, each organizational model requires separate infrastructure or custom integration, increasing operational cost and limiting platform scalability.
+
+The platform needs a tenant model with unlimited hierarchy depth (configurable advisory threshold), visibility barriers for self-managed tenants, and consistent tenant context propagation across all services.
+
+### 1.3 Goals (Business Outcomes)
+
+- Provide one stable contract for tenant hierarchy, type enforcement, barrier semantics, and user group coordination (via RG).
+- Support both managed (no barrier — delegated administration) and self-managed (with barrier — independent operation) tenant modes within the same tenant tree, covering use-cases from MSP service delivery to autonomous subsidiaries.
+- Enable unlimited hierarchy depth with a configurable advisory threshold (default: 10 levels) for diverse organizational models.
```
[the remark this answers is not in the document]
diffora: Infinite hierarchy is useless from business perspective - maybe, but we don't know which level of hierarchy is required, so it's the platform owner's choice

modules/system/accounts/docs/PRD.md
```diff
@@ -0,0 +1,1568 @@
...
+- Enable end-to-end tenant-context validation to achieve p95 ≤ 5ms on every API call.
+- Maintain zero cross-tenant data leaks, verified by automated security tests.
+- Provide a pluggable IdP integration contract for user lifecycle operations.
+
+**Success criteria:**
+
+| Metric | Baseline | Target | Timeframe |
+|--------|----------|--------|-----------|
+| Tenant onboarding effort | Manual multi-step provisioning per organizational model | Single API call — tenant create with type and mode | Module GA |
```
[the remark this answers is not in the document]
diffora: The API doesn't hide business differences — it parameterizes them via tenant_type (reseller, customer, MSP…) and self_managed (managed vs autonomous). GTS enforces which combinations are valid per deployment. What's unified is the structural plumbing (create node, link IdP, set barrier), not the business semantics.

Domain-specific onboarding logic lives in consuming layers (portals, partner mgmt) that call AM after their own workflows. AM stays out of the way.

modules/system/accounts/docs/PRD.md
```diff
@@ -0,0 +1,1568 @@
...
+- Being an IdP implementation — HAS consumes IdP, not replaces it.
+- User-tenant reassignment (moving a user between tenants) — deferred from v1; requires cross-platform coordination (RG membership migration, resource ownership transfer, AuthZ cache invalidation, session revocation).
+
+### 1.5 Glossary
+
+> Platform-wide terms (Tenant, Subject Tenant, Context Tenant, Resource Tenant) are defined canonically in the [Authorization Core Terms](../../../../docs/arch/authorization/DESIGN.md#core-terms) and the [Tenant Model](../../../../docs/arch/authorization/TENANT_MODEL.md). This glossary re-states them briefly for self-containedness and adds HAS-specific terms.
+
+| Term | Definition |
+|------|------------|
+| Tenant | Logical organizational entity representing a company, division, team, or individual. The fundamental unit of multi-tenancy isolation. |
```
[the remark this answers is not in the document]
diffora: The account management module name was discussed: https://github.com/cyberfabric/cyberfabric-core/pull/1373#discussion_r3021788870

modules/system/accounts/docs/PRD.md
```diff
@@ -0,0 +1,1568 @@
...
+- [9. Acceptance Criteria](#9-acceptance-criteria)
+- [10. Dependencies](#10-dependencies)
+- [11. Assumptions](#11-assumptions)
+- [12. Risks](#12-risks)
+- [13. Open Questions](#13-open-questions)
+- [14. Traceability](#14-traceability)
+
+<!-- /toc -->
+
+> **Abbreviation**: Hierarchical Account Server = **HAS**. Used throughout this document.
```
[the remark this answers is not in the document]
diffora: Outdated.

modules/system/account-management/docs/ADR/0004-cpt-cf-accounts-adr-resource-group-tenant-hierarchy-source.md
```diff
@@ -0,0 +1,150 @@
+---
+status: rejected
```
entropyshift, line 2: should be `accepted`?

modules/system/account-management/docs/ADR/0003-cpt-cf-accounts-adr-conversion-approval.md
```diff
@@ -0,0 +1,111 @@
+---
+status: superseded
```
entropyshift: should be `accepted`?

modules/system/account-management/docs/ADR/0002-cpt-cf-accounts-adr-metadata-inheritance.md
```diff
@@ -0,0 +1,109 @@
+---
+status: superseded
```
entropyshift: should be `accepted`?

modules/system/account-management/docs/ADR/0004-cpt-cf-accounts-adr-resource-group-tenant-hierarchy-source.md
diffora: No, it's rejected. It's about using a resource group as a storage for tenants.

modules/system/account-management/docs/ADR/0003-cpt-cf-accounts-adr-conversion-approval.md
diffora: Please look at **Superseded By** block.

modules/system/account-management/docs/ADR/0002-cpt-cf-accounts-adr-metadata-inheritance.md
diffora: Please look at **Superseded By** block.

modules/system/account-management/docs/DESIGN.md
```diff
@@ -0,0 +1,1934 @@
...
+
+IdP integration uses the CyberFabric gateway + plugin pattern, analogous to AuthN Resolver (see `cpt-cf-accounts-adr-idp-contract-separation`). AM defines an `IdpProviderPluginClient` trait for tenant and user administrative operations (tenant provisioning/deprovisioning, user provision/deprovision, impersonation, tenant-scoped query). The plugin is discovered via GTS types-registry and resolved through `ClientHub`. The platform ships a default provider plugin; vendors substitute their own implementation behind the same trait. The IdP provider plugin is intentionally separate from the AuthN Resolver plugin — the two target different concerns (admin operations vs hot-path token validation) with different performance profiles and protocols. The contract is one-directional: AM calls IdP, IdP does not call AM.
```
aviator5, line 51: > `IdpProviderPluginClient` trait for tenant and user administrative operations (tenant provisioning/deprovisioning, user provision/deprovision, impersonation, tenant-scoped query)

This phrase confuses a bit: I thought AM itself will be responsible for tenant-management, and IdP - only for identities.

modules/system/account-management/docs/DESIGN.md
```diff
@@ -0,0 +1,1934 @@
...
+
+User group management is handled by the [Resource Group](../../resource-group/docs/PRD.md) module. AM registers a dedicated Resource Group type for user groups during module initialization; consumers call `ResourceGroupClient` directly for group lifecycle, membership, and hierarchy operations.
```
aviator5, line 53: > AM registers a dedicated Resource Group type for user groups during module initialization

Need to define GTS schema for this

modules/system/account-management/docs/DESIGN.md
```diff
@@ -0,0 +1,1934 @@
...
+
+#### System Context
+
+```mermaid
+graph LR
+    Admin["Platform / Tenant<br>Administrator"] -->|REST API| AM["Account<br>Management"]
+    AM -->|provision/deprovision<br>users & tenants| IdP["IdP Provider<br>Plugin"]
+    AM -->|type validation<br>& schema queries| GTS["GTS Types<br>Registry"]
+    AM -->|user-group type<br>registration & cleanup| RG["Resource<br>Group"]
+    AM -->|resource pre-check<br>on delete| RMS["Resource<br>Management"]
```
aviator5: Could you please clarify what "Resource Management" is here?

modules/system/account-management/docs/DESIGN.md
```diff
@@ -0,0 +1,1934 @@
...
+| `cpt-cf-accounts-fr-idp-tenant-deprovision` | Background hard-deletion job calls `IdpProviderPluginClient::deprovision_tenant` for every hard-deleted tenant. Provider implementations clean up tenant-scoped IdP resources, guided by tenant type traits such as `idp_provisioning`. Providers **MUST NOT** silently no-op on mutating operations — unsupported deprovisioning **MUST** fail with `idp_unsupported_operation`. Failure retries rather than skips. |
+| `cpt-cf-accounts-fr-idp-user-provision` | `IdpProviderPluginClient::create_user` with tenant scope binding and resolved tenant metadata (for IdP context resolution, e.g., effective Keycloak realm). |
+| `cpt-cf-accounts-fr-idp-user-deprovision` | `IdpProviderPluginClient::delete_user` with session revocation. |
+| `cpt-cf-accounts-fr-idp-user-query` | `IdpProviderPluginClient::list_users` with tenant filter; supports optional user-ID filter for single-user lookups. |
+| `cpt-cf-accounts-fr-user-group-rg-type` | `AccountsModule` idempotently registers the user-group Resource Group type during module initialization, with `allowed_memberships` including the platform IdP user resource type (`gts.z.system.idp.user.v1~`). |
```
aviator5: `gts.z.system.idp.user.v1~` -> `gts.x.core.idp.user.v1~`?

modules/system/account-management/docs/ADR/0002-cpt-cf-accounts-adr-metadata-inheritance.md
aviator5: Sorry, but do we really need to commit an ADR that is superseded in the same PR?
For me, it's unclear whether should I review this file or not?

modules/system/account-management/docs/ADR/0004-cpt-cf-accounts-adr-barrier-aware-metadata-inheritance.md
```diff
@@ -0,0 +1,109 @@
...
+- [More Information](#more-information)
+- [Traceability](#traceability)
+
+<!-- /toc -->
+
+**ID**: `cpt-cf-accounts-adr-barrier-aware-metadata-inheritance`
+
+## Context and Problem Statement
+
+AM already chose read-time walk-up resolution for inherited tenant metadata, but the historical ADR left one consequence stated incorrectly: whether the `/resolved` walk may cross a self-managed barrier. The canonical PRD and DESIGN both require a self-managed tenant to be metadata-independent from ancestors above its barrier, so we need an explicit decision for barrier semantics within the walk-up algorithm.
```
aviator5: Could you please provide a concrete use-cases when tenants' metadata inheritance is needed? In my view, typical metadata is contact details and it shouldn't be inherited.

modules/system/account-management/docs/DESIGN.md
```diff
@@ -0,0 +1,1934 @@
...
+
+**Base type traits** (defined in `x-gts-traits-schema`):
+
+| Trait | Type | Default | Description |
+|-------|------|---------|-------------|
+| `can_be_root` | boolean | `false` | Whether tenants of this type can be root tenants (`parent_id = null`) |
+| `allowed_parent_types` | string[] | `[]` | GTS instance identifiers of tenant types allowed as parent. Empty array means no nesting allowed. |
+| `idp_provisioning` | boolean | `false` | Whether tenants of this type typically require dedicated IdP-side resources. AM still invokes `provision_tenant` / `deprovision_tenant`; provider implementations may use this trait to decide whether to create dedicated resources or reuse shared ones. Mutating IdP methods **MUST NOT** silently no-op — providers that do not support a required operation **MUST** return `idp_unsupported_operation`. |
```
aviator5, line 303: Why for some tenants we might want not to do IdP provisioning?

modules/system/account-management/docs/DESIGN.md
```diff
@@ -0,0 +1,1934 @@
...
+
+Derived type schemas resolve their behavioral traits via `x-gts-traits` (per [GTS spec §9.7](https://github.com/GlobalTypeSystem/gts-spec?tab=readme-ov-file#97---schema-traits-x-gts-traits-schema--x-gts-traits)). Properties not specified fall back to defaults from the base type's `x-gts-traits-schema`.
+
+**Example — Cloud Hosting Deployment:**
+
+| GTS Schema ID (chained, stored) | Description | `x-gts-traits` |
+|---------------------------------|-------------|----------------|
+| `gts.x.accounts.tenants.tenant_type.v1~x.accounts.tenants.provider.v1~` | Platform operator; root tenant | `can_be_root: true`, `idp_provisioning: true` |
```
aviator5: 1. Please check other modules, seems like we use singular form in GTS IDs (accounts -> account, tenants -> tenant, etc.).
2. If Account Management is a system module, I believe it's worth to have GTS ID like `gts.x.core.am.tenant_type.v1~`. In my view "am" short form is ok since we already have `gts.x.core.rg.type.v1~` ("rg" - Resource Group). @Artifizer, @diffora, @entropyshift thoughts?

modules/system/account-management/docs/DESIGN.md
```diff
@@ -0,0 +1,1934 @@
...
...he same domain method, which resolves the caller's side from the endpoint and determines initiation vs approval from `ConversionRequest` state. Conversion request cancellation (`cancel_conversion`) transitions a pending request to `cancelled` status; either the initiator (withdraw) or counterparty (reject) can cancel from their own tenant scope. IdP user operations (provision, deprovision, query, impersonation) via `IdpProviderPluginClient`. Children query with pagination and status filtering. Root tenant listing (Platform Administrator only, `parent_id IS NULL` query). Inbound conversion-request listing for parent tenants (surfaces pending requests from barrier-isolated children without bypassing the barrier). Hard deletion scheduling after retention period; hard-deletion job processes tenants in leaf-first order (`depth DESC`) to satisfy the `parent_id` FK without cascading deletes. Provisioning reaper job cleans stale `provisioning` tenants (older than configurable timeout, default: 5 minutes) by calling `deprovision_tenant` (idempotent) and deleting the row. Deterministic error mapping for all failure paths. Platform request audit logging captures all state-changing operations.
```
aviator5: > a background provisioning reaper either retries finalization or compensates

From my perspective we need to do only compensation, not finalization. Otherwise it may lead to unexpected behavior for the AM clients.

aviator5: Suggest splitting this paragraph into sections - hard to read...

modules/system/account-management/docs/DESIGN.md
```diff
@@ -0,0 +1,1934 @@
...
+- **Technology**: REST / OpenAPI
+- **Planned location**: `modules/system/account-management/account-management/src/api/rest/`
+
+**Planned**: Machine-readable OpenAPI specification at `modules/system/account-management/schemas/openapi/accounts-v1.yaml`, to be generated from OperationBuilder annotations at build time.
+
+For the consolidated AuthN/AuthZ mapping of all endpoints to PolicyEnforcer actions, see the Resource Types and Actions table in Security Architecture (Section 4).
+
+**Endpoints Overview**:
+
+| Method | Path | Description | Stability |
```
aviator5, line 512: Sorry, but what what is the goal of the Stability column?

modules/system/account-management/docs/DESIGN.md
```diff
@@ -0,0 +1,1934 @@
...
+|--------|------|-------------|-----------|
+| `POST` | `/api/accounts/v1/tenants` | Create tenant (root or child based on `parent_id`) | stable |
+| `GET` | `/api/accounts/v1/tenants/{id}` | Get tenant by ID | stable |
+| `PATCH` | `/api/accounts/v1/tenants/{id}` | Update tenant (name, status) | stable |
+| `DELETE` | `/api/accounts/v1/tenants/{id}` | Soft-delete tenant | stable |
+| `GET` | `/api/accounts/v1/tenants/{id}/children` | List direct children (paginated, filterable) | stable |
+| `GET` | `/api/accounts/v1/tenants?parent_id=null` | List root tenants (Platform Administrator only) | stable |
```
aviator5: In what cases we need to read all roots? What token will be used for this?

modules/system/account-management/docs/DESIGN.md
```diff
@@ -0,0 +1,1934 @@
...
+| `GET` | `/api/accounts/v1/tenants/{id}/conversion-requests` | List inbound conversion requests for parent tenant | stable |
+| `POST` | `/api/accounts/v1/tenants/{id}/convert` | Initiate or approve mode conversion (child-scoped) | stable |
+| `POST` | `/api/accounts/v1/tenants/{parent_id}/children/{child_id}/convert` | Initiate or approve mode conversion (parent-scoped) | stable |
+| `DELETE` | `/api/accounts/v1/tenants/{id}/convert` | Cancel pending conversion request (child-scoped) | stable |
+| `DELETE` | `/api/accounts/v1/tenants/{parent_id}/children/{child_id}/convert` | Cancel pending conversion request (parent-scoped) | stable |
+| `POST` | `/api/accounts/v1/tenants/{id}/impersonate` | Obtain impersonation token for managed child | stable |
```
aviator5, line 524: If AM starts issuing tokens, it will partially take IdP responsibility here. Also, I don't understand what subject_id will be in those tokens.

modules/system/account-management/docs/ADR/0003-cpt-cf-accounts-adr-conversion-approval.md
diffora: will merge the ADRs

modules/system/account-management/docs/ADR/0002-cpt-cf-accounts-adr-metadata-inheritance.md
diffora: will merge the ADRs

modules/system/account-management/docs/DESIGN.md
```diff
@@ -0,0 +1,1934 @@
...
+| Method | Path | Description | Stability |
+|--------|------|-------------|-----------|
+| `POST` | `/api/accounts/v1/tenants` | Create tenant (root or child based on `parent_id`) | stable |
+| `GET` | `/api/accounts/v1/tenants/{id}` | Get tenant by ID | stable |
+| `PATCH` | `/api/accounts/v1/tenants/{id}` | Update tenant (name, status) | stable |
+| `DELETE` | `/api/accounts/v1/tenants/{id}` | Soft-delete tenant | stable |
+| `GET` | `/api/accounts/v1/tenants/{id}/children` | List direct children (paginated, filterable) | stable |
+| `GET` | `/api/accounts/v1/tenants?parent_id=null` | List root tenants (Platform Administrator only) | stable |
+| `GET` | `/api/accounts/v1/tenants/{id}/conversion-requests` | List inbound conversion requests for parent tenant | stable |
+| `POST` | `/api/accounts/v1/tenants/{id}/convert` | Initiate or approve mode conversion (child-scoped) | stable |
```
aviator5, line 520: Did you consider an option to set/unset self-managed flag in the `PATCH /api/accounts/v1/tenants/{id}` endpoint instead of `POST /api/accounts/v1/tenants/{id}/convert` and `POST /api/accounts/v1/tenants/{parent_id}/children/{child_id}/convert`?

modules/system/account-management/docs/DESIGN.md
diffora: AM decides which tenant a user belongs to.
The IdP stores the user account, so we need IdpProviderPluginClient to persist and read that tenant linkage in the IdP.
So we have two options:
1) Option A (current) to proxy the user provisioning in AM and link the user.
2) Option B - have a method to link the user with the tenant
Any suggestions here?

modules/system/account-management/docs/DESIGN.md
diffora: Resource Management - is an RMS system with a resource graph and lifecycle.
You are right -we don't need it here because resource relations are managed through resource groups (will delete it).

modules/system/account-management/docs/DESIGN.md
diffora: will add

modules/system/account-management/docs/DESIGN.md
diffora: Agree. Fixed.

modules/system/account-management/docs/ADR/0004-cpt-cf-accounts-adr-barrier-aware-metadata-inheritance.md
diffora: **branding**
A provider or reseller tenant defines logo, colors, portal title, support URL once.
All managed child tenants inherit that branding by default.
Only exceptional tenants override it.

**contacts**
Parent defines default support, security, billing, or escalation contacts.
Child tenants inherit those contacts until they have their own local team.
This avoids copying the same contact set into every tenant.

**billing-address**
A corporate parent owns billing centrally.
Subsidiary or department tenants inherit the parent billing address unless invoiced separately.
When a child becomes independently billed, it overrides locally.

**IdP context metadata such as keycloak-realm**
If a child tenant does not have its own IdP realm metadata, it can inherit the parent realm, so user operations still resolve to a valid IdP context.
If the child gets its own realm, it overrides the inherited value.

modules/system/account-management/docs/DESIGN.md
diffora: Some tenants do not need their own IdP space.

Examples:

- several tenants share one realm/directory
- a tenant is only for billing/org structure, not identity isolation
- creating a separate realm per tenant is too expensive or unnecessary

So idp_provisioning = false means:
don’t create dedicated IdP resources for this tenant; use shared or existing IdP context instead.

modules/system/account-management/docs/DESIGN.md
diffora: Yes, "gts.x.core.am" sounds good

modules/system/account-management/docs/DESIGN.md
diffora: done.

diffora: Changed to compensation only.

modules/system/account-management/docs/DESIGN.md
diffora: Typical meaning:

- **Stable**: safe to implement against as a v1 contract
- **Planned** or similar: expected, but details may still change
- **Internal** or **Experimental**: not a normal public compatibility promise

So the column is mainly a consumer-facing contract signal, not a runtime behavior property.

modules/system/account-management/docs/DESIGN.md
diffora: GET /tenants?parent_id=null is for forest-level administration, not normal tenant work.

Concrete cases:
- auditing existing top-level tenant trees
- capacity / inventory views across independent tenant trees
- Platform administrator to view the all tenants on the platform

The token is just a normal authenticated IdP access token for a principal that has a platform-level Platform Administrator grant

modules/system/account-management/docs/DESIGN.md
diffora: We need the approval workflow only for self-managed -> managed, because that change removes the barrier and expands the parent’s visibility and operational access to the child tenant. That is not a normal field update, it is a trust-boundary change between two independent admin scopes.

A unilateral change would let one side restore cross-tenant control without the other side’s consent. The approval workflow gives explicit bilateral consent, a durable pending state, expiry, and an audit trail. By contrast, managed -> self-managed does not need approval because it only reduces access and increases isolation.

See: 0003-cpt-cf-accounts-adr-conversion-approval.md

modules/system/account-management/docs/DESIGN.md
diffora: 1) AM must decide impersonation because it owns the tenant relationship and barrier rules. The IdP only issues the token.
So:

- AM checks whether parent -> child impersonation is allowed
- IdP creates the impersonation token

If all of this moved into the IdP, we would have to duplicate AM’s tenant model there and tightly couple the platform to a specific IdP.

2) Suggested claim layout:

sub: real user ID of the parent admin
subject_tenant_id: home tenant of that admin
context_tenant_id: impersonated child tenant

**I agree that impersonation is a hard thing and maybe can be moved to the next versions. What do you think?**

modules/system/account-management/docs/DESIGN.md
aviator5: After discussion with @diffora and @nonameffh we decided to change tenant model topology (`docs/arch/authorization/TENANT_MODEL.md`) from the forest to the single-root. There are two main reasons for this:
1) We will be able to simplify multiple places in the Account Management module.
2) If there is a forest, we will have to have multiple client credentials (for each root tenant) and manage them for proper [S2S Authentication](https://github.com/cyberfabric/cyberfabric-core/blob/main/docs/arch/authorization/DESIGN.md#s2s-authentication-service-to-service) in case of tenant-scoped operations.

@diffora please update PRD and design docs for the Account Management accordingly. I will update `docs/arch/authorization/TENANT_MODEL.md` document in #1527.

@Artifizer, @entropyshift FYI

modules/system/account-management/docs/DESIGN.md
aviator5: I'm ok with Option A. Option B will lead to multiple questions.

My concern was about the phrase itself: from my perspective, please correct me if I'm wrong, but`IdpProviderPluginClient` will not do the tenant provisioning/deprovisioning (it will be on the Account Management module itself).

diffora: create tenant
  -> save tenant as provisioning
  -> call IDP via `IdpProviderPluginClient::provision_tenant(...)`
  -> save returned IDP metadata
  -> set tenant status=active
  -> return created

It's a subcall in the creating process.

modules/system/account-management/docs/DESIGN.md
diffora: done.

modules/system/account-management/docs/DESIGN.md
aviator5: Yes, I would suggest having "Open Questions" section and a new item about impersonation there.

modules/system/account-management/docs/DESIGN.md
aviator5: Actually, I'm not fully convinced that `managed` -> `self_managed` should be unilateral.
This transition is not only an isolation change, but also a significant governance / operating model change between the child and the parent. Right now the design treats `managed` -> `self_managed` as immediate, while `self_managed` -> `managed` already has a request-based lifecycle with approval, expiry, and cancellation.

Would it be cleaner and more consistent to model both directions via the same request concept? For example, any mode transition could be represented as a first-class request, initiated by either side and then approved by the counterparty. That would give a single mental model for all mode changes and make the approval semantics more explicit.

modules/system/account-management/docs/DESIGN.md
```diff
@@ -0,0 +1,1948 @@
...
+|--------|------|-------------|-----------|
+| `POST` | `/api/accounts/v1/tenants` | Create child tenant | stable |
+| `GET` | `/api/accounts/v1/tenants/{id}` | Get tenant by ID | stable |
+| `PATCH` | `/api/accounts/v1/tenants/{id}` | Update tenant (name, status) | stable |
+| `DELETE` | `/api/accounts/v1/tenants/{id}` | Soft-delete tenant | stable |
+| `GET` | `/api/accounts/v1/tenants/{id}/children` | List direct children (paginated, filterable) | stable |
+| `GET` | `/api/accounts/v1/tenants/{id}/conversion-requests` | List inbound conversion requests for parent tenant | stable |
+| `POST` | `/api/accounts/v1/tenants/{id}/convert` | Initiate or approve mode conversion (child-scoped) | stable |
+| `POST` | `/api/accounts/v1/tenants/{parent_id}/children/{child_id}/convert` | Initiate or approve mode conversion (parent-scoped) | stable |
+| `DELETE` | `/api/accounts/v1/tenants/{id}/convert` | Cancel pending conversion request (child-scoped) | stable |
```
aviator5, line 522: Just an idea.

The API surface feels unnecessarily action-oriented and a bit overloaded.
`convert` is too generic and does not communicate that this is specifically a tenant mode transition between `managed` and `self_managed`. At the same time, the design already introduces a dedicated stateful request entity with its own lifecycle, so the external API could probably reflect that model more directly.

Would it be cleaner to expose this as a first-class resource instead of multiple `/convert` endpoints, for example:
+ GET /api/account-management/v1/tenant-mode-change-requests
+ POST /api/account-management/v1/tenant-mode-change-requests
+ GET /api/account-management/v1/tenant-mode-change-requests/{request_id}
+ PATCH /api/account-management/v1/tenant-mode-change-requests/{request_id}
+ DELETE /api/account-management/v1/tenant-mode-change-requests/{request_id}

This would make the API more self-descriptive and align it better with the underlying domain model.

Partially correlates with another suggestion in https://github.com/cyberfabric/cyberfabric-core/pull/1373/changes#r3096239216.

modules/system/account-management/docs/DESIGN.md
```diff
@@ -0,0 +1,1948 @@
...
+- **Interfaces**: `cpt-cf-accounts-interface-tenant-mgmt-api`
+- **Contracts**: `cpt-cf-accounts-contract-tenant-resolver`, `cpt-cf-accounts-contract-authz-resolver`
+- **Technology**: REST / OpenAPI
+- **Planned location**: `modules/system/account-management/account-management/src/api/rest/`
+
+**Planned**: Machine-readable OpenAPI specification at `modules/system/account-management/docs/schemas/accounts-v1.yaml`, to be generated from OperationBuilder annotations at build time.
+
+For the consolidated AuthN/AuthZ mapping of all endpoints to PolicyEnforcer actions, see the Resource Types and Actions table in Security Architecture (Section 4).
+
+**Endpoints Overview**:
```
aviator5, line 510: Set of `/api/accounts/v1/tenants/{id}/metadata` endpoints is missing.

modules/system/account-management/docs/DESIGN.md
```diff
@@ -0,0 +1,1948 @@
...
+
+- [ ] `p2` - **ID**: `cpt-cf-accounts-component-metadata-service`
+
+##### Why this component exists
+
+Manages extensible tenant metadata with GTS-schema validation and per-schema inheritance resolution.
+
+##### Responsibility scope
+
+Metadata CRUD (create, read, update, delete) for any GTS-registered metadata kind. Schema validation against GTS at write time. Inheritance resolution: for `inherit` schemas, walks the tenant hierarchy upward to find the nearest ancestor value; for `override-only` schemas, returns the tenant's own value or empty. Cascade deletion of metadata entries when a tenant is deprovisioned.
```
aviator5, line 463: A few questions on metadata modeling.

What is the exact format of kind? Is it a human-readable logical key (for example branding, contacts) or a GTS identifier?

How is kind mapped to the GTS schema/type used for validation?

Also, should we introduce a common base GTS type for tenant metadata in addition to per-kind schemas, so that all metadata items share an explicit common contract? At minimum, it seems we need a common `inheritance_policy` field there (from my perspective it's better if we prefer enum instead of the bool flag, but it's up to you), because inheritance semantics are clearly modeled as a per-schema contract rather than an implementation detail.

modules/system/account-management/docs/DESIGN.md
diffora: Please review proposal https://github.com/diffora/cyberfabric-core/blob/a305101f34ae4b64252fbc0dc31686959b51a891/modules/system/account-management/docs/proposals/mode-conversion-approval-v2.md

modules/system/account-management/docs/DESIGN.md
diffora: Good catch — the docs are inconsistent today, and you're right that inheritance belongs in the schema contract, not in MetadataService prose. Here's what we'll change:

1. kind is a GTS schema identifier, not a logical key. The DB already stores it as tenant_metadata.schema_id TEXT (example: gts.z.cf.metadata.branding.v1~). Only the URL param is misleadingly named. We'll rename {kind} → {schema_id} in the REST API so URL and DB column match 1:1, and add an explicit note: "schema_id is a full chained GTS schema identifier; URL-encode the ~."

2. Mapping made explicit. schema_id is the single key into both the GTS types registry (for schema + trait lookup) and the tenant_metadata table (UNIQUE (tenant_id, schema_id)). MetadataService flow: fetch schema by schema_id → 404 not_found if unregistered → validate body → upsert. This will be written out in the Metadata Behavioral Notes.

3. New base GTS type for tenant metadata — agreed. Mirrors the tenant_type.v1.schema.json pattern. New schema gts.x.core.am.tenant_metadata.v1~ with x-gts-traits-schema declaring inheritance_policy as an enum (override_only | inherit), default override_only:

"inheritance_policy": {
  "type": "string",
  "enum": ["override_only", "inherit"],
  "default": "override_only"
}
Enum over bool for future-proofing (merge, readonly, computed are plausible later) and consistency with existing enum-style traits like barrier_mode. MetadataService resolves the policy from the registered schema's traits — no side config, same path tenant-type traits already use for idp_provisioning.

What do you think?

modules/system/account-management/docs/PRD.md
```diff
@@ -0,0 +1,1770 @@
...
+
+- [ ] `p1` - **ID**: `cpt-cf-accounts-fr-bootstrap-ordering`
+
+**Actors**: `cpt-cf-accounts-actor-platform-admin`, `cpt-cf-accounts-actor-idp`
+
+The system **MUST** wait for the IdP to be available before completing bootstrap, retrying with backoff, and failing after a configurable timeout if the IdP is not ready.
+
+- **Rationale**: The root tenant cannot be fully operational without its associated IdP; proceeding without it would leave the platform in an inconsistent state.
+
+### 5.2 Tenant Hierarchy Management
```
aviator5, line 387: I could not find an explicit design for tenant re-parenting ("tenant move" operation).

The current PRD describes tenant creation, update, delete, mode conversion, and metadata inheritance, but it does not seem to define whether an existing tenant (or subtree) can be moved under another parent.

This feels like an important gap, because tenant move is a widely used operation pattern.

If tenant move is intentionally out of scope for v1, I think it would be better to state that explicitly. Otherwise, this likely needs its own design section / ADR.

aviator5: Ok, I see "Hierarchy reparenting (moving a tenant subtree to a new parent)" in the "Known Limitations & Technical Debt" section in the DESIGN.md. @diffora, as we discussed let's keep this item for the future design, but please provide some internal details how tenants' move could be implemented in the AM and what pitfalls are here (conversion requests cancelling, prohibiting tenants move between IdP provisioned tenants, etc.).

modules/system/account-management/docs/DESIGN.md
aviator5: Agreed, makes sense

modules/system/account-management/docs/DESIGN.md
aviator5: I like the minimalism of Option A, but I’m not sure the singular /tenants/{id}/conversion sub-resource is the best long-term API shape here. The underlying model is already a first-class request entity with its own id, status, TTL, audit trail, and historical rows, so exposing it as a collection resource may align better with the actual domain model. Right now the proposal keeps the transport surface minimal, but the handler semantics are still fairly state-driven: the same PUT may create a request, approve an existing one, or become a same-side idempotent no-op depending on current server state.

I wonder if a more explicit resource-oriented API would age better, for example:

+ GET /api/accounts/v1/tenants/{id}/conversions
+ POST /api/accounts/v1/tenants/{id}/conversions
+ GET /api/accounts/v1/tenants/{id}/conversions/{request_id}
+ PATCH /api/accounts/v1/tenants/{id}/conversions/{request_id}
+ GET /api/accounts/v1/tenants/{id}/child-conversions

With this shape, POST would always create a new conversion request, while PATCH would drive the lifecycle explicitly (approved, rejected, cancelled, etc.) instead of overloading PUT / DELETE with multiple meanings. That would make the process model more explicit and easier to evolve later.

The main reason I’m raising this is future extensibility. Even if we do not want to fully design those features now, this flow already looks like something that may later need management capabilities: initiator comment/message on creation, editing that comment while the request is pending, reject/cancel explanation from the counterparty, historical browsing, filtering, and richer audit visibility. A collection-based request API seems more naturally prepared for that than a singular stateful /conversion endpoint.

modules/system/account-management/docs/DESIGN.md
diffora: added.

modules/system/account-management/docs/DESIGN.md
diffora: moved.

modules/system/account-management/docs/DESIGN.md
diffora: The final API surface for conversion requests is:

- `GET /api/account-management/v1/tenants/{tenant_id}/conversions`  
  Lists conversion requests visible from the child tenant's own scope.

- `POST /api/account-management/v1/tenants/{tenant_id}/conversions`  
  Creates a new `ConversionRequest` initiated from the child side.

- `GET /api/account-management/v1/tenants/{tenant_id}/conversions/{request_id}`  
  Returns one conversion request in child scope.

- `PATCH /api/account-management/v1/tenants/{tenant_id}/conversions/{request_id}`  
  Applies an allowed lifecycle transition on an existing child-scope request, such as `approved`, `rejected`, or `cancelled`.

- `GET /api/account-management/v1/tenants/{tenant_id}/child-conversions`  
  Lists conversion requests targeting the tenant's direct children, using the restricted parent-scope discovery view.

- `POST /api/account-management/v1/tenants/{tenant_id}/child-conversions`  
  Creates a new `ConversionRequest` initiated by the parent for one direct child. Here `{tenant_id}` is the parent tenant scope; the target child is provided separately as `child_tenant_id`.

- `GET /api/account-management/v1/tenants/{tenant_id}/child-conversions/{request_id}`  
  Returns one conversion request in parent scope. Here `{tenant_id}` is still the parent tenant whose child-conversion view is being used.

- `PATCH /api/account-management/v1/tenants/{tenant_id}/child-conversions/{request_id}`  
  Applies an allowed lifecycle transition on an existing parent-scope request.

This keeps the API resource-oriented around durable `ConversionRequest` entities, while preserving separate child-scope and parent-scope collections because those two views have different authorization and data-exposure rules.

aviator5: looks good

modules/system/account-management/docs/DESIGN.md
diffora: resolved here https://github.com/cyberfabric/cyberfabric-core/pull/1373#discussion_r3110449586

modules/system/account-management/docs/PRD.md
diffora: Added

modules/system/account-management/docs/migration.sql
```diff
@@ -0,0 +1,146 @@
...
+
+-- ── Tenants ──────────────────────────────────────────────────────────────────
+
+CREATE TABLE tenants (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    parent_id UUID NULL,
+    name TEXT NOT NULL CHECK (length(name) BETWEEN 1 AND 255),
+    status TEXT NOT NULL CHECK (status IN ('provisioning', 'active', 'suspended', 'deleted')),
+    self_managed BOOLEAN NOT NULL DEFAULT FALSE,
+    tenant_type_schema_id gts_type_path NOT NULL,
```
aviator5: Should `tenant_type_schema_id` be UUID as well?
As we discussed, it may be calculated as UUIDv5(GTS_ID) too

modules/system/account-management/docs/schemas/authz_vocabulary.v1.schema.json
```diff
@@ -0,0 +1,124 @@
+{
+  "$id": "gts://gts.x.core.am.authz_vocabulary.v1~",
```
aviator5: Suggest removing this scheme and just mentioning permissions (permission = resource_type + action) that will be checked in the Account Management module.

From my perspective, it's worth to have `gts.x.core.authz.permission.v1.~` base GTS type with the following fields in the scheme: `resource_type` (string, GTS ref), `action` (string) and `display_name` (string).

Example of such permissions:
+ `gts.x.core.authz.permission.v1.~x.core.am.tenant_create.v1`
+ `gts.x.core.authz.permission.v1.~x.core.am.user_delete.v1`
+ etc.

It will allow us implementing canonical "AuthZ Management" module that will play PDP (Policy Decision Point, via AuthZResolver contract) and PAP (Policy Administration Point, roles).

I've a created a dedicated issue for this: #1563 (on me).

modules/system/account-management/docs/account-management-v1.yaml
aviator5, line 1: [nitpick] Suggest moving openapi contract outside of the schemas dir - looks like it's supposed to contain only GTS schemas. if not, why database schema is not here (`migration.sql`).

modules/system/account-management/docs/PRD.md
```diff
@@ -0,0 +1,1893 @@
...
+| Cross-tenant data leaks | 0 AM-specific automated isolation scenarios wired into the gate today | Zero leaks observed; 100% of the planned AM isolation scenarios pass | Pre-GA security gate |
+| Hierarchy depth coverage | Fixed shallow hierarchies (2–3 levels) | Unlimited depth with configurable advisory threshold (default: 10); validated with at least 3 distinct type topologies including hierarchies beyond the default threshold | Pre-GA integration test gate |
+| Conforming IdP provider implementations | 0 provider implementations validated against the AM contract suite | ≥ 2 provider implementations pass the AM tenant/user lifecycle contract suite without breaking the public API contract | Pre-GA integration gate |
+
+### 1.4 Non-goals
+
+- User authentication flows (covered by IAM PRD).
+- Authorization policy evaluation or SQL predicate generation (covered by AuthZ Resolver).
+- Resource provisioning and lifecycle (covered by Resource Management System PRD).
```
aviator5: Sorry, but it's still unclear what Resource Management System is and why it's mentioned here.

diffora: I will delete this line.

modules/system/account-management/docs/account-management-v1.yaml
diffora: Agree. Moved.

modules/system/account-management/docs/schemas/authz_vocabulary.v1.schema.json
diffora: I will remove that file.

modules/system/account-management/docs/migration.sql
diffora: Yes, Agree.

aviator5 (approved):

modules/system/account-management/docs/migration.sql
```diff
@@ -0,0 +1,146 @@
...
+
+-- Reference DDL for the Account Management source-of-truth schema.
+-- This file is intentionally documentation-first: implementation migrations may
+-- express the same logical schema through ModKit/SeaORM migration code.
+
+CREATE EXTENSION IF NOT EXISTS pgcrypto;
+
+-- ── GTS type path domain ─────────────────────────────────────────────────────
+-- GTS type identifier: single or chained schema id, always ending with ~.
+CREATE DOMAIN gts_type_path AS TEXT
```
aviator5: Can be removed..

diffora: 100%

modules/system/accounts/docs/PRD.md
diffora: Added paragraph.

Tenant hierarchy is needed because many customers operate as parent/child organizations rather than flat accounts. It lets the platform model delegated administration, governance boundaries, inherited defaults, and future billing/reporting flows in a way that matches real customer structure.

modules/system/accounts/docs/PRD.md
diffora: The PRD now says:

- AM has no hard structural depth cap in the data model 
- the default advisory threshold remains 10
- deeper production support must be backed by approved performance benchmarks on the deployment profile, including dependent resolver paths

modules/system/accounts/docs/PRD.md
diffora: Added:

Here, a deployment means one installed/running Cyber Fabric environment with a single AM root tenant and one configured tenant-type topology. Different organizational shapes are modeled as tenant hierarchies inside that deployment, not as separate platform installations.

modules/system/accounts/docs/PRD.md
```diff
@@ -0,0 +1,1568 @@
...
+
+### 1.4 Non-goals
+
+- User authentication flows (covered by IAM PRD).
+- Authorization policy evaluation or SQL predicate generation (covered by AuthZ Resolver).
+- Resource provisioning and lifecycle (covered by Resource Management System PRD).
+- Being an IdP implementation — HAS consumes IdP, not replaces it.
+- User-tenant reassignment (moving a user between tenants) — deferred from v1; requires cross-platform coordination (RG membership migration, resource ownership transfer, AuthZ cache invalidation, session revocation).
+
+### 1.5 Glossary
```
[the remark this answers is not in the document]
diffora: Agreed this was ambiguous in the earlier wording. AM is not a commercial-entity management domain; it is the platform tenant-management domain. It owns tenant hierarchy, tenant mode, tenant-scoped admin metadata, and IdP-backed admin lifecycle coordination, while subscriptions, plans, entitlements, and billing remain outside its scope. The "onboarding" wording is meant as platform tenant setup, not commercial customer onboarding.

entropyshift (approved):

nonameffh (approved):

Artifizer (approved):

PR #1395, opened 2026-04-01
fix(oagw): handle missing SecurityContext for CORS preflight requests

ffedoroff:
Body:

  ## Summary

  - CORS preflight (OPTIONS) requests to `/oagw/v1/proxy/{alias}/*` return 500 instead of the expected CORS response
  - Root cause: `authn_middleware` correctly skips authentication for CORS preflights (no SecurityContext injected), but `proxy_handler` unconditionally
  requires `Extension<SecurityContext>` — axum fails to extract it and returns 500
  - Fix: detect CORS preflight in `proxy_handler` before requiring SecurityContext and handle it directly, bypassing the authenticated proxy pipeline

  ## Test plan

  - [ ] Run `E2E_AUTH_TOKEN=e2e-token-tenant-a make e2e-local`
  - [ ] All 5 previously failing tests in `testing/e2e/modules/oagw/test_cors.py` now pass:
    - `test_cors_preflight_allowed_origin`
    - `test_cors_preflight_rejected_origin`
    - `test_cors_preflight_rejected_method`
    - `test_cors_preflight_rejected_header`
    - `test_cors_preflight_with_credentials`
  - [ ] No regression in other oagw tests (198 → 203 passed)

### DIFF
--- a/config/e2e-local.yaml
+++ b/config/e2e-local.yaml
@@ -111,7 +111,7 @@ modules:
           in_flight: 64
 
       # Authentication Configuration
-      auth_disabled: true
+      auth_disabled: false
       require_auth_by_default: true
 
   # --- Tenant Resolver example (gateway + plugins) ---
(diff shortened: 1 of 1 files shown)


Artifizer:
Cancelling, fixed in another PR from asmith

PR #1403, opened 2026-04-02
feat(dylint): add de1101 lint for tests in separate files

ffedoroff:
## PR chain

This is **PR 1c of 8** in the resource-group feature train.

All 8 PRs together deliver the `resource-group` module. This PR adds a lint rule that enforces the test file layout convention the resource-group module is built with — landing it first means CI will guard that convention going forward.

**Chain order:**
1a. #1401 docs/rg-1a-docs
1c. #1403 **→ feat/rg-1c-dylint** *(this PR)*
1d. #1404 feat/rg-1d-odata
1e. #1405 feat/rg-1e-db-security
2.  #1406 feat/rg-2-authz
3.  #1407 feat/rg-3-resource-group
4.  #1408 test/rg-4-tests

---

## Why this PR exists

The resource-group module separates all tests into `tests/` integration test binaries and `*_tests.rs` inline files, rather than using `#[cfg(test)] mod tests { ... }` blocks inside production source files. This separation matters because:

1. **Compilation speed** — inline test code is compiled on every `cargo build --tests`, even in CI steps that only need the binary. Separate files are only compiled when tests actually run.
2. **Reviewability** — a source file with a large inline test block is harder to review; keeping tests in dedicated files makes the production logic easy to read.
3. **dylint enforcement** — without a lint rule, the convention silently erodes. This PR adds `de1101` to catch inline `#[cfg(test)] mod tests` blocks in `src/` files before they are merged.

The UI test cases in this PR use the resource-group module's own file layout as the canonical "correct" example, which means the lint ships with proof that it does not fire on the module it was designed for.

---

## Summary

Add `de1101_tests_in_separate_files` dylint rule: emits a warning when a `#[cfg(test)]` module is defined inline in a non-test source file.

---

## Rule behavior

**Rejected** (lint fires):
```rust
// src/domain/foo.rs
#[cfg(test)]
mod tests {
    #[test]
    fn it_works() { ... }
}
```

**Allowed** (lint passes):
```
src/domain/foo.rs          ← production code only
src/domain/foo_tests.rs    ← tests, included via mod declaration
tests/domain_unit_test.rs  ← integration test binary
```

---

## UI test cases

| Snapshot | Scenario | Expected outcome |
|----------|----------|-----------------|
| `resource_group_inline_case` | Inline `mod tests` in a domain file | Lint fires |
| `resource_group_out_of_line_test_module` | Out-of-line test module still inside `src/` | Lint fires |
| `resource_group_separate_test_file_tests` | Correct layout: `*_tests.rs` file | Lint passes |
| `other_module_inline_tests` | Inline tests in an unrelated module | Lint fires |

---

## Test plan

- [ ] `cargo test` in `dylint_lints/` passes — `.stderr` snapshot files match exactly
- [ ] Lint does not fire on `tests/` directory or `*_tests.rs` include files
- [ ] Lint fires on inline `#[cfg(test)] mod tests` blocks in `src/` files
- [ ] `lint_utils` shared helpers compile cleanly

## Summary by CodeRabbit

* **New Features**
  * Added DE1101 lint to enforce resource-group tests live in separate files.

* **Documentation**
  * Added detailed DE1101 docs with examples and guidance.

* **Tests**
  * Added UI test cases and expected outputs covering allowed and disallowed patterns.

* **Chores**
  * Enabled the new lint crate in the workspace and re-enabled an existing lint.

* **Bug Fixes**
  * Minor change to cursor-related error message formatting.


MikeFalcon77:
please fix dynlint CI

PR #1466, opened 2026-04-08
bugfix: compile error when no-default-features are specified

Bechma:
## Release Notes

* **Refactor**
  * Telemetry initialization is now exclusively available when OpenTelemetry support is enabled, removing inaccessible code paths from the public API.

### DIFF
--- a/libs/modkit/src/telemetry/init.rs
+++ b/libs/modkit/src/telemetry/init.rs
@@ -190,17 +190,6 @@ pub fn init_tracing(
     Ok(otel_layer)
 }
 
-/// No-op when the `otel` feature is disabled.
-///
-/// # Errors
-/// Always returns an error indicating the feature is disabled.
-#[cfg(not(feature = "otel"))]
-pub fn init_tracing(
-    _otel_cfg: &super::config::OpenTelemetryConfig,
-) -> anyhow::Result<crate::bootstrap::host::logging::OtelLayer> {
-    Err(anyhow::anyhow!("otel feature is disabled"))
-}
-
 #[cfg(feature = "otel")]
 pub(crate) fn build_headers_from_cfg_and_env(
     exporter: Option<&crate::telemetry::config::Exporter>,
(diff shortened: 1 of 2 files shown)


MikeFalcon77:
## Suggestion: keep the no-op stub, fix the root cause

The compile error happens because the stub references `crate::bootstrap::host::logging::OtelLayer` — and the `bootstrap` module is behind its own feature gate. When `--no-default-features` disables both `otel` and `bootstrap`, the path doesn't resolve.

Removing the stub works, but it pushes `#[cfg(feature = "otel")]` onto every call site. The callers in `oop.rs` and `run.rs` already have their own `#[cfg]` guards because of this — ideally they shouldn't need them.

**Better approach**: move the `OtelLayer` type alias into `telemetry` (which is always available) and make the stub return `Ok(())` instead of `Err`:

```rust
// telemetry/mod.rs (or telemetry/init.rs)
#[cfg(feature = "otel")]
pub type OtelLayer = tracing_opentelemetry::OpenTelemetryLayer<
    tracing_subscriber::Registry,
    opentelemetry_sdk::trace::Tracer,
>;
#[cfg(not(feature = "otel"))]
pub type OtelLayer = ();
```

```rust
#[cfg(not(feature = "otel"))]
pub fn init_tracing(_cfg: &OpenTelemetryConfig) -> anyhow::Result<OtelLayer> {
    Ok(())  // no-op, not an error — disabled feature is normal state
}
```

This way:
- The public API is identical with or without `otel`
- Call sites don't need `#[cfg]` guards — `init_tracing` always exists and always returns `Ok`
- No cross-feature dependency on `bootstrap`
- Follows the standard Rust pattern used by `tracing`, `metrics`, etc. — same signature, feature switches the implementation

The same applies to `init_metrics_provider` no-op stub which currently returns `Err` — it should return `Ok(())` for consistency.

MikeFalcon77:
Discussed internally. We will leave it with feature gate

MikeFalcon77 (approved):

PR #1481, opened 2026-04-09
tests: api_contracts should not generate and compare api.json

joseph-cx:
## Summary by CodeRabbit

* **Chores**
  * CI workflow now runs only for changes to the API spec or the workflow itself.
  * Removed the spec freshness verification phase and its related build/compare steps.
  * Breaking-change detection retained and simplified: compare base vs PR spec and fail only on detected breaking changes unless explicitly acknowledged.
  * Final workflow summary simplified to report base-spec presence and breaking-change results.

### DIFF
--- a/.github/workflows/api_contracts.yml
+++ b/.github/workflows/api_contracts.yml
@@ -1,24 +1,15 @@
-# This workflow enforces two API contract checks on every PR:
+# This workflow enforces API contract checks on PRs that modify the API spec
+# (or this workflow file):
 #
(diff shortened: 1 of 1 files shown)


joseph-cx:
@coderabbitai review

Artifizer:
It seems like this PR is not needed because original logic is fine - because we can't rely on the fact the API spec was properly updated in a user PR, that's what we need to check exactly.

Artifizer (changes requested):
see my comments above

PR #1492, opened 2026-04-12
feat(resource-group): SDK and module contracts

ffedoroff:
## Summary
- Resource Group SDK crate with API traits, models, error types, and OData filter definitions
- Module skeleton with domain layer (repo traits, validation, error types) and infrastructure (Sea-ORM entities, storage)
- Workspace and Cargo.toml wiring for the new `resource-group-sdk` and `resource-group` crates

## PR Train 🚂
Merge in order:
1. you are here — #1492

Next PR train https://github.com/cyberfabric/cyberfabric-core/pull/1552

## Summary by CodeRabbit

* **New Features**
  * Added Resource Group system and SDK: manage types and groups (create/read/update/delete), group hierarchy depth queries, membership management, OData filtering, and JSON Schema metadata validation.

* **Documentation**
  * Updated design, PRD, and feature docs to use the revised GTS type-path prefix and aligned examples/tests.

* **Chores**
  * Registered new workspace packages and updated project ignore rules.

* **Security**
  * Adjusted advisory handling: one advisory removed from the ignore list.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,4 @@
+# Updated: 2026-04-16 by Constructor Tech
 [workspace.package]
 edition = "2024"
 license = "Apache-2.0"
@@ -53,6 +54,8 @@ members = [
     "modules/system/grpc-hub",
     "modules/system/nodes-registry/nodes-registry",
     "modules/system/nodes-registry/nodes-registry-sdk",
+    "modules/system/resource-group/resource-group-sdk",
+    "modules/system/resource-group/resource-group",
     "modules/system/module-orchestrator",
     "examples/modkit/users-info/users-info-sdk",
     "examples/modkit/users-info/users-info",
@@ -239,6 +242,7 @@ cf-system-sdk-directory = { version = "0.1.34", path = "libs/system-sdks/sdks/di
 types-registry-sdk = { package = "cf-types-registry-sdk", version = "0.1.5", path = "modules/system/types-registry/types-registry-sdk" }
 tenant-resolver-sdk = { package = "cf-tenant-resolver-sdk", version = "0.2.17", path = "modules/system/tenant-resolver/tenant-resolver-sdk" }
 authz-resolver-sdk = { package = "cf-authz-resolver-sdk", version = "0.2.18", path = "modules/system/authz-resolver/authz-resolver-sdk" }
+resource-group-sdk = { package = "cf-resource-group-sdk", version = "0.1.0", path = "modules/system/resource-group/resource-group-sdk" }
 oagw-sdk = { package = "cf-oagw-sdk", version = "0.4.0", path = "modules/system/oagw/oagw-sdk" }
 
 # credstore
@@ -253,7 +257,7 @@ grpc_hub = { package = "cf-grpc-hub", version = "0.1.19", path = "modules/system
 # Serialization
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-serde-saphyr = "0.0.16"
+serde-saphyr = "0.0.22"
 secrecy = { version = "0.10", features = ["serde"] }
 serde_urlencoded = "0.7"
 schemars = { version = "1.2", features = ["derive", "uuid1"] }
@@ -408,7 +412,7 @@ regex = "1.10"
 glob = "0.3"
 colored = "3.1"
 supports-color = "3"
-enable-ansi-support = "0.2"
+enable-ansi-support = "0.3"
 
 # gRPC support
 tonic = { version = "0.14", features = ["transport"] }
@@ -451,7 +455,7 @@ hostname = "0.4"
 starship-battery = "0.10"
 local-ip-address = "0.6"
 machine-uid = "0.5"
-nvml-wrapper = "0.11"
+nvml-wrapper = "0.12"
 
 # Public Suffix List
 psl = "2"
@@ -466,7 +470,7 @@ strsim = "0.11"
 
 # OData parsing
 peg = "0.8"
-chrono-tz = "0.9"
+chrono-tz = "0.10"
 
 # Procedural macros
 proc-macro2 = "1.0"
@@ -484,7 +488,7 @@ proc-macro-error2 = "2.0"
 
 # Testing utilities
 trybuild = "1"
-criterion = { version = "0.5", default-features = false }
+criterion = { version = "0.8", default-features = false }
 testcontainers = { version = "0.27", default-features = false, features = ["aws-lc-rs"] }
 testcontainers-modules = { version = "0.15", default-features = false, features = ["aws-lc-rs", "postgres", "mysql"] }
 httpmock = "0.8"
@@ -523,7 +527,7 @@ nanoid = "0.4"
 tokio-metrics = { version = "0.4", features = ["rt"] }
 
 # GTS dependencies
-jsonschema = "0.40"
+jsonschema = { version = "0.40", default-features = false }
 walkdir = "2.5"
 shellexpand = "3.1"
 
--- a/modules/system/resource-group/resource-group-sdk/Cargo.toml
+++ b/modules/system/resource-group/resource-group-sdk/Cargo.toml
@@ -1,0 +1,44 @@
+# Created: 2026-04-16 by Constructor Tech
+[package]
+name = "cf-resource-group-sdk"
+version = "0.1.0"
+publish = false
+edition.workspace = true
+license.workspace = true
+authors.workspace = true
+description = "SDK for resource-group module: API trait, types, and error definitions"
+
+[lints]
+workspace = true
+
+[features]
+default = ["odata"]
+odata = [
+    "dep:modkit-odata-macros",
+    "dep:modkit-sdk",
+]
+
+[dependencies]
+# Core dependencies for API trait
+thiserror = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+async-trait = { workspace = true }
+
+# OData macros (for #[derive(ODataFilterable)])
+modkit-odata-macros = { workspace = true, optional = true }
+
+# UUID support
+uuid = { workspace = true }
+
+# Security context for API methods
+modkit-security = { workspace = true }
+
+# SDK utilities (typed OData queries)
+modkit-sdk = { workspace = true, optional = true }
+
+# OData support for pagination and query types (required for API trait)
+modkit-odata = { workspace = true }
+
+# GTS type system — ID parsing and validation
+gts = { workspace = true }
--- a/modules/system/resource-group/resource-group-sdk/Cargo.toml
+++ b/modules/system/resource-group/resource-group-sdk/Cargo.toml
@@ -42,3 +42,12 @@ modkit-odata = { workspace = true }
 
 # GTS type system — ID parsing and validation
 gts = { workspace = true }
+
+# GTS macros — struct_to_gts_schema for GTS schema definitions
+gts-macros = { workspace = true }
+
+# JSON Schema (required by struct_to_gts_schema)
+schemars = { workspace = true }
+
+# ModKit — BaseModkitPluginV1 for plugin spec base type
+modkit = { workspace = true }
--- a/modules/system/resource-group/resource-group-sdk/src/api.rs
+++ b/modules/system/resource-group/resource-group-sdk/src/api.rs
@@ -1,0 +1,181 @@
+// Created: 2026-04-16 by Constructor Tech
+// @cpt-dod:cpt-cf-resource-group-dod-sdk-foundation-sdk-traits:p1
+//! SDK trait contracts for the resource-group module.
+
+use async_trait::async_trait;
+use modkit_security::SecurityContext;
+
+use modkit_odata::{ODataQuery, Page};
+use uuid::Uuid;
+
+use crate::error::ResourceGroupError;
+use crate::models::{
+    CreateGroupRequest, CreateTypeRequest, PatchGroupRequest, ResourceGroup,
+    ResourceGroupMembership, ResourceGroupType, ResourceGroupWithDepth, UpdateGroupRequest,
+    UpdateTypeRequest,
+};
+
+/// Client trait for resource-group type management.
+///
+/// Consumers obtain this from `ClientHub`:
+/// ```ignore
+/// let client = hub.get::<dyn ResourceGroupClient>()?;
+/// let rg_type = client.get_type(&ctx, "gts.cf.core.rg.type.v1~...").await?;
+/// ```
+#[async_trait]
+pub trait ResourceGroupClient: Send + Sync {
+    // -- Type lifecycle --
+
+    /// Create a new GTS type definition.
+    async fn create_type(
+        &self,
+        ctx: &SecurityContext,
+        request: CreateTypeRequest,
+    ) -> Result<ResourceGroupType, ResourceGroupError>;
+
+    /// Get a GTS type definition by its code (GTS type path).
+    async fn get_type(
+        &self,
+        ctx: &SecurityContext,
+        code: &str,
+    ) -> Result<ResourceGroupType, ResourceGroupError>;
+
+    /// List GTS type definitions with `OData` filtering and cursor-based pagination.
+    async fn list_types(
+        &self,
+        ctx: &SecurityContext,
+        query: &ODataQuery,
+    ) -> Result<Page<ResourceGroupType>, ResourceGroupError>;
+
+    /// Update a GTS type definition (full replacement).
+    async fn update_type(
+        &self,
+        ctx: &SecurityContext,
+        code: &str,
+        request: UpdateTypeRequest,
+    ) -> Result<ResourceGroupType, ResourceGroupError>;
+
+    /// Delete a GTS type definition. Fails if groups of this type exist.
+    async fn delete_type(
+        &self,
+        ctx: &SecurityContext,
+        code: &str,
+    ) -> Result<(), ResourceGroupError>;
+
+    // -- Group lifecycle --
+
+    /// Create a new resource group.
+    async fn create_group(
+        &self,
+        ctx: &SecurityContext,
+        request: CreateGroupRequest,
+    ) -> Result<ResourceGroup, ResourceGroupError>;
+
+    /// Get a resource group by ID.
+    async fn get_group(
+        &self,
+        ctx: &SecurityContext,
+        id: Uuid,
+    ) -> Result<ResourceGroup, ResourceGroupError>;
+
+    /// List resource groups with `OData` filtering and cursor-based pagination.
+    async fn list_groups(
+        &self,
+        ctx: &SecurityContext,
+        query: &ODataQuery,
+    ) -> Result<Page<ResourceGroup>, ResourceGroupError>;
+
+    /// Update a resource group (full replacement).
+    async fn update_group(
+        &self,
+        ctx: &SecurityContext,
+        id: Uuid,
+        request: UpdateGroupRequest,
+    ) -> Result<ResourceGroup, ResourceGroupError>;
+
+    /// Patch a resource group (partial update via PATCH).
+    async fn patch_group(
+        &self,
+        ctx: &SecurityContext,
+        id: Uuid,
+        request: PatchGroupRequest,
+    ) -> Result<ResourceGroup, ResourceGroupError>;
+
+    /// Delete a resource group.
+    /// When `force` is false, fails with `ConflictActiveReferences` if child groups
+    /// or memberships exist. When `force` is true, recursively deletes the entire
+    /// subtree (all descendants and their memberships).
+    async fn delete_group(
+        &self,
+        ctx: &SecurityContext,
+        id: Uuid,
+        force: bool,
+    ) -> Result<(), ResourceGroupError>;
+
+    /// List group hierarchy with relative depth from a reference group.
+    async fn list_group_depth(
+        &self,
+        ctx: &SecurityContext,
+        group_id: Uuid,
+        query: &ODataQuery,
+    ) -> Result<Page<ResourceGroupWithDepth>, ResourceGroupError>;
+
+    // -- Membership lifecycle --
+
+    /// Add a membership link between a resource and a group.
+    async fn add_membership(
+        &self,
+        ctx: &SecurityContext,
+        group_id: Uuid,
+        resource_type: &str,
+        resource_id: &str,
+    ) -> Result<ResourceGroupMembership, ResourceGroupError>;
+
+    /// Remove a membership link.
+    async fn remove_membership(
+        &self,
+        ctx: &SecurityContext,
+        group_id: Uuid,
+        resource_type: &str,
+        resource_id: &str,
+    ) -> Result<(), ResourceGroupError>;
+
+    /// List memberships with `OData` filtering and cursor-based pagination.
+    async fn list_memberships(
+        &self,
+        ctx: &SecurityContext,
+        query: &ODataQuery,
+    ) -> Result<Page<ResourceGroupMembership>, ResourceGroupError>;
+}
+
+// @cpt-dod:cpt-cf-resource-group-dod-integration-auth-read-service:p1
+/// Narrow read-only trait for hierarchy data, used by `AuthZ` plugin.
+///
+/// This trait provides the integration read port that external consumers
+/// (such as the `AuthZ` plugin) use to query group hierarchy data without
+/// depending on the full `ResourceGroupClient`.
+#[async_trait]
+pub trait ResourceGroupReadHierarchy: Send + Sync {
+    /// List group hierarchy with depth for a given group.
+    async fn list_group_depth(
+        &self,
+        ctx: &SecurityContext,
+        group_id: Uuid,
+        query: &ODataQuery,
+    ) -> Result<Page<ResourceGroupWithDepth>, ResourceGroupError>;
+}
+
+// @cpt-flow:cpt-cf-resource-group-flow-integration-auth-plugin-routing:p1
+/// Extended read trait for vendor-specific plugin gateway routing.
+///
+/// Extends `ResourceGroupReadHierarchy` with membership listing for
+/// plugin consumers that need both hierarchy and membership data.
+#[async_trait]
+pub trait ResourceGroupReadPluginClient: ResourceGroupReadHierarchy {
+    /// List memberships for plugin consumers.
+    async fn list_memberships(
+        &self,
+        ctx: &SecurityContext,
+        query: &ODataQuery,
+    ) -> Result<Page<ResourceGroupMembership>, ResourceGroupError>;
+}
--- a/modules/system/resource-group/resource-group-sdk/src/api.rs
+++ b/modules/system/resource-group/resource-group-sdk/src/api.rs
@@ -10,9 +10,8 @@ use uuid::Uuid;
 
 use crate::error::ResourceGroupError;
 use crate::models::{
-    CreateGroupRequest, CreateTypeRequest, PatchGroupRequest, ResourceGroup,
-    ResourceGroupMembership, ResourceGroupType, ResourceGroupWithDepth, UpdateGroupRequest,
-    UpdateTypeRequest,
+    CreateGroupRequest, CreateTypeRequest, ResourceGroup, ResourceGroupMembership,
+    ResourceGroupType, ResourceGroupWithDepth, UpdateGroupRequest, UpdateTypeRequest,
 };
 
 /// Client trait for resource-group type management.
@@ -93,27 +92,24 @@ pub trait ResourceGroupClient: Send + Sync {
         request: UpdateGroupRequest,
     ) -> Result<ResourceGroup, ResourceGroupError>;
 
-    /// Patch a resource group (partial update via PATCH).
-    async fn patch_group(
-        &self,
-        ctx: &SecurityContext,
-        id: Uuid,
-        request: PatchGroupRequest,
-    ) -> Result<ResourceGroup, ResourceGroupError>;
-
     /// Delete a resource group.
-    /// When `force` is false, fails with `ConflictActiveReferences` if child groups
-    /// or memberships exist. When `force` is true, recursively deletes the entire
-    /// subtree (all descendants and their memberships).
-    async fn delete_group(
+    ///
+    /// SDK deletes are non-cascade: the call fails with
+    /// `ConflictActiveReferences` if the group has child groups or memberships.
+    /// Cascade deletion is only available through the REST surface.
+    async fn delete_group(&self, ctx: &SecurityContext, id: Uuid)
+    -> Result<(), ResourceGroupError>;
+
+    /// Get descendants of a reference group (depth >= 0).
+    async fn get_group_descendants(
         &self,
         ctx: &SecurityContext,
-        id: Uuid,
-        force: bool,
-    ) -> Result<(), ResourceGroupError>;
+        group_id: Uuid,
+        query: &ODataQuery,
+    ) -> Result<Page<ResourceGroupWithDepth>, ResourceGroupError>;
 
-    /// List group hierarchy with relative depth from a reference group.
-    async fn list_group_depth(
+    /// Get ancestors of a reference group (depth <= 0).
+    async fn get_group_ancestors(
         &self,
         ctx: &SecurityContext,
         group_id: Uuid,
@@ -156,8 +152,16 @@ pub trait ResourceGroupClient: Send + Sync {
 /// depending on the full `ResourceGroupClient`.
 #[async_trait]
 pub trait ResourceGroupReadHierarchy: Send + Sync {
-    /// List group hierarchy with depth for a given group.
-    async fn list_group_depth(
+    /// Get descendants of a reference group (depth >= 0).
+    async fn get_group_descendants(
+        &self,
+        ctx: &SecurityContext,
+        group_id: Uuid,
+        query: &ODataQuery,
+    ) -> Result<Page<ResourceGroupWithDepth>, ResourceGroupError>;
+
+    /// Get ancestors of a reference group (depth <= 0).
+    async fn get_group_ancestors(
         &self,
         ctx: &SecurityContext,
         group_id: Uuid,
--- a/modules/system/resource-group/resource-group-sdk/src/api.rs
+++ b/modules/system/resource-group/resource-group-sdk/src/api.rs
@@ -164,18 +164,3 @@ pub trait ResourceGroupReadHierarchy: Send + Sync {
         query: &ODataQuery,
     ) -> Result<Page<ResourceGroupWithDepth>, ResourceGroupError>;
 }
-
-// @cpt-flow:cpt-cf-resource-group-flow-integration-auth-plugin-routing:p1
-/// Extended read trait for vendor-specific plugin gateway routing.
-///
-/// Extends `ResourceGroupReadHierarchy` with membership listing for
-/// plugin consumers that need both hierarchy and membership data.
-#[async_trait]
-pub trait ResourceGroupReadPluginClient: ResourceGroupReadHierarchy {
-    /// List memberships for plugin consumers.
-    async fn list_memberships(
-        &self,
-        ctx: &SecurityContext,
-        query: &ODataQuery,
-    ) -> Result<Page<ResourceGroupMembership>, ResourceGroupError>;
-}
--- a/modules/system/resource-group/resource-group-sdk/src/error.rs
+++ b/modules/system/resource-group/resource-group-sdk/src/error.rs
@@ -1,0 +1,151 @@
+// Created: 2026-04-16 by Constructor Tech
+// @cpt-dod:cpt-cf-resource-group-dod-sdk-foundation-sdk-errors:p1
+//! Public error types for the resource-group module.
+//!
+//! These errors are safe to expose to other modules and consumers.
+
+use thiserror::Error;
+
+/// Errors that can be returned by the `ResourceGroupClient`.
+#[derive(Error, Debug, Clone)]
+pub enum ResourceGroupError {
+    /// Resource with the specified identifier was not found.
+    #[error("Resource not found: {code}")]
+    NotFound { code: String },
+
+    /// A resource with the specified code already exists.
+    #[error("Resource already exists: {code}")]
+    TypeAlreadyExists { code: String },
+
+    /// Validation error with the provided data.
+    #[error("Validation error: {message}")]
+    Validation { message: String },
+
+    /// Removing allowed parents or disabling root placement would break
+    /// existing group hierarchy relationships.
+    #[error("Allowed parents violation: {message}")]
+    AllowedParentsViolation { message: String },
+
+    /// Cannot delete a type because groups of this type still exist.
+    #[error("Active references exist: {message}")]
+    ConflictActiveReferences { message: String },
+
+    /// A generic conflict (e.g. a concurrency or state conflict not related to references).
+    #[error("Conflict: {message}")]
+    Conflict { message: String },
+
+    /// Parent type is not allowed by the type's `allowed_parents` configuration.
+    #[error("Invalid parent type: {message}")]
+    InvalidParentType { message: String },
+
+    /// A cycle would be created in the group hierarchy.
+    #[error("Cycle detected: {message}")]
+    CycleDetected { message: String },
+
+    /// A configured limit (depth, width, etc.) would be exceeded.
+    #[error("Limit violation: {message}")]
+    LimitViolation { message: String },
+
+    /// Tenant scope incompatibility.
+    #[error("Tenant incompatibility: {message}")]
+    TenantIncompatibility { message: String },
+
+    /// Service is temporarily unavailable.
+    #[error("Service unavailable: {message}")]
+    ServiceUnavailable { message: String },
+
+    /// Access was denied by the authorization policy (PDP denial).
+    #[error("Access denied")]
+    AccessDenied,
+
+    /// An internal error occurred.
+    #[error("Internal error")]
+    Internal,
+}
+
+impl ResourceGroupError {
+    /// Create a `NotFound` error.
+    pub fn not_found(code: impl Into<String>) -> Self {
+        Self::NotFound { code: code.into() }
+    }
+
+    /// Create a `TypeAlreadyExists` error.
+    pub fn type_already_exists(code: impl Into<String>) -> Self {
+        Self::TypeAlreadyExists { code: code.into() }
+    }
+
+    /// Create a `Validation` error.
+    pub fn validation(message: impl Into<String>) -> Self {
+        Self::Validation {
+            message: message.into(),
+        }
+    }
+
+    /// Create an `AllowedParentsViolation` error.
+    pub fn allowed_parents_violation(message: impl Into<String>) -> Self {
+        Self::AllowedParentsViolation {
+            message: message.into(),
+        }
+    }
+
+    /// Create a `ConflictActiveReferences` error.
+    pub fn conflict_active_references(message: impl Into<String>) -> Self {
+        Self::ConflictActiveReferences {
+            message: message.into(),
+        }
+    }
+
+    /// Create a generic `Conflict` error.
+    pub fn conflict(message: impl Into<String>) -> Self {
+        Self::Conflict {
+            message: message.into(),
+        }
+    }
+
+    /// Create an `InvalidParentType` error.
+    pub fn invalid_parent_type(message: impl Into<String>) -> Self {
+        Self::InvalidParentType {
+            message: message.into(),
+        }
+    }
+
+    /// Create a `CycleDetected` error.
+    pub fn cycle_detected(message: impl Into<String>) -> Self {
+        Self::CycleDetected {
+            message: message.into(),
+        }
+    }
+
+    /// Create a `LimitViolation` error.
+    pub fn limit_violation(message: impl Into<String>) -> Self {
+        Self::LimitViolation {
+            message: message.into(),
+        }
+    }
+
+    /// Create a `TenantIncompatibility` error.
+    pub fn tenant_incompatibility(message: impl Into<String>) -> Self {
+        Self::TenantIncompatibility {
+            message: message.into(),
+        }
+    }
+
+    /// Create a `ServiceUnavailable` error.
+    pub fn service_unavailable(message: impl Into<String>) -> Self {
+        Self::ServiceUnavailable {
+            message: message.into(),
+        }
+    }
+
+    /// Create an `AccessDenied` error.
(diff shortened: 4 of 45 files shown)


Cargo.toml
```diff
@@ -238,6 +240,7 @@ cf-system-sdk-directory = { version = "0.1.34", path = "libs/system-sdks/sdks/di
 types-registry-sdk = { package = "cf-types-registry-sdk", version = "0.1.5", path = "modules/system/types-registry/types-registry-sdk" }
 tenant-resolver-sdk = { package = "cf-tenant-resolver-sdk", version = "0.2.17", path = "modules/system/tenant-resolver/tenant-resolver-sdk" }
 authz-resolver-sdk = { package = "cf-authz-resolver-sdk", version = "0.2.18", path = "modules/system/authz-resolver/authz-resolver-sdk" }
+resource-group-sdk = { package = "cf-resource-group-sdk", version = "0.5.10", path = "modules/system/resource-group/resource-group-sdk" }
```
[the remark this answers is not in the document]
ffedoroff: Acknowledged. The workspace version is set intentionally — the SDK crate uses `workspace = true` so the actual version resolved comes from the workspace root. No mismatch at build time.

modules/system/resource-group/resource-group-sdk/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+[lints]
+workspace = true
+
+[features]
+default = ["odata"]
+odata = [
+    "dep:modkit-odata-macros",
+    "dep:modkit-sdk",
+    "dep:modkit-odata",
+]
```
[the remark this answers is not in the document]
ffedoroff: The `modkit-odata` dependency is always enabled when building the resource-group crate (it's a required dependency of the parent crate). The optional flag in the SDK is for downstream consumers who may not need OData support. Won't change.

modules/system/resource-group/resource-group-sdk/src/error.rs
```diff
@@ -0,0 +1,129 @@
...
+//! These errors are safe to expose to other modules and consumers.
+
+use thiserror::Error;
+
+/// Errors that can be returned by the `ResourceGroupClient`.
+#[derive(Error, Debug, Clone)]
+pub enum ResourceGroupError {
+    /// Resource with the specified code was not found.
+    #[error("Type not found: {code}")]
+    NotFound { code: String },
```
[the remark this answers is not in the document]
ffedoroff: Fixed in this push — error messages now use `"Resource not found: {code}"` and `"Resource already exists: {code}"` since the enum is shared across groups, types, and memberships.

modules/system/resource-group/resource-group-sdk/src/models.rs
```diff
@@ -0,0 +1,572 @@
...
+    /// Create a new `GtsTypePath` from a raw string, applying validation.
+    ///
+    /// # Errors
+    /// Returns an error if the string is empty, exceeds 255 characters,
+    /// or does not match the GTS type path format.
+    pub fn new(raw: impl Into<String>) -> Result<Self, String> {
+        // @cpt-begin:cpt-cf-resource-group-algo-sdk-foundation-validate-gts-type-path:p1:inst-gts-val-2
+        let raw = raw.into();
+        let s = raw.trim().to_lowercase();
+        // @cpt-end:cpt-cf-resource-group-algo-sdk-foundation-validate-gts-type-path:p1:inst-gts-val-2
```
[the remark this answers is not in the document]
ffedoroff: The validation logic is intentional — `GtsTypePath` follows the same pattern as other path types in the codebase. The domain-layer validator adds module-specific constraints on top.

modules/system/resource-group/resource-group-sdk/src/models.rs
```diff
@@ -0,0 +1,572 @@
...
+/// `null` clear the value. Fields with a value update it.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[allow(clippy::option_option)]
+pub struct PatchGroupRequest {
+    pub name: Option<String>,
+    #[serde(default, deserialize_with = "option_option::deserialize")]
+    pub parent_id: Option<Option<Uuid>>,
+    #[serde(default, deserialize_with = "option_option::deserialize")]
+    pub metadata: Option<Option<serde_json::Value>>,
```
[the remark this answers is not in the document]
ffedoroff: The stored schema merging strategy is by design: we keep system fields in the outer object and user-defined schema under a known key. This is consistent with how other modules handle JSON Schema composition.

modules/system/resource-group/resource-group-sdk/src/odata/groups.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    fn kind(&self) -> FieldKind {
+        match self {
+            // Type is SMALLINT in DB; filter values are resolved from string to
+            // integer at the persistence boundary before OData processing.
+            Self::Type => FieldKind::I64,
+            Self::Name => FieldKind::String,
+            Self::HierarchyParentId | Self::Id => FieldKind::Uuid,
+        }
```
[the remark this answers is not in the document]
ffedoroff: False positive: `GroupFilterField::Type` already uses `FieldKind::String` in the current code.

modules/system/resource-group/resource-group-sdk/src/odata/hierarchy.rs
```diff
@@ -0,0 +1,53 @@
...
+        match self {
+            Self::HierarchyDepth => "hierarchy/depth",
+            Self::Type => "type",
+        }
+    }
+
+    fn kind(&self) -> FieldKind {
+        match self {
+            Self::HierarchyDepth | Self::Type => FieldKind::I64,
+        }
```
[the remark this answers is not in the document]
ffedoroff: False positive: `HierarchyFilterField::Type` already uses `FieldKind::String`.

modules/system/resource-group/resource-group/Cargo.toml
```diff
@@ -0,0 +1,66 @@
+[package]
```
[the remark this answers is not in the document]
ffedoroff: The `cypilot_path` is configured at the workspace root level, not per-crate. This is consistent with all other modules.

modules/system/resource-group/resource-group/src/domain/error.rs
```diff
@@ -0,0 +1,209 @@
...
+    /// Returns `true` if this error represents a serialization failure
+    /// (SQLSTATE `40001`) that is safe to retry.
+    ///
+    /// Covers `PostgreSQL` `SERIALIZABLE` conflicts and `MySQL`/`MariaDB` deadlocks.
+    #[must_use]
+    pub fn is_serialization_failure(&self) -> bool {
+        match self {
+            DomainError::Database { message } => {
+                message.contains("40001") || message.contains("could not serialize access")
+            }
```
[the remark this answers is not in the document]
ffedoroff: We only target PostgreSQL in production. MySQL/MariaDB deadlock codes are intentionally not covered. The SQLSTATE `40001` handles Postgres serialization failures which is what we need.

modules/system/resource-group/resource-group/src/domain/error.rs
```diff
@@ -0,0 +1,209 @@
...
+            DomainError::CycleDetected { message } => ResourceGroupError::cycle_detected(message),
+            DomainError::LimitViolation { message } => ResourceGroupError::limit_violation(message),
+            DomainError::AllowedParentsViolation { message } => {
+                ResourceGroupError::allowed_parents_violation(message)
+            }
+            DomainError::ConflictActiveReferences { message }
+            | DomainError::Conflict { message } => {
+                ResourceGroupError::conflict_active_references(message)
+            }
```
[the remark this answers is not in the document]
ffedoroff: The error hierarchy follows the same pattern as other modules in the codebase — domain errors are intentionally granular, SDK errors are intentionally coarse. This is by design.

modules/system/resource-group/resource-group/src/domain/repo.rs
```diff
@@ -0,0 +1,287 @@
...
+use modkit_security::AccessScope;
+use resource_group_sdk::models::{
+    ResourceGroup, ResourceGroupMembership, ResourceGroupType, ResourceGroupWithDepth,
+};
+use uuid::Uuid;
+
+use crate::domain::error::DomainError;
+use crate::infra::storage::entity::{
+    gts_type, resource_group as rg_entity, resource_group_membership as membership_entity,
+};
```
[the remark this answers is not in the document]
ffedoroff: Valid architectural observation. In practice, all modules in this codebase follow the same pattern where domain traits reference entity types from the infra layer. This is a known pragmatic trade-off. Refactoring to pure domain models would be a cross-cutting change.

modules/system/resource-group/resource-group/src/domain/repo.rs
```diff
@@ -0,0 +1,287 @@
...
+#[allow(clippy::too_many_arguments)]
+pub trait GroupRepositoryTrait: Send + Sync + 'static {
+    // -- Read operations --
+
+    async fn find_by_id<C: DBRunner>(
+        &self,
+        db: &C,
+        scope: &AccessScope,
+        id: Uuid,
+    ) -> Result<Option<ResourceGroup>, DomainError>;
```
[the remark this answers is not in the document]
ffedoroff: Not changing. The service layer gates access via `access_scope()` before every repo call — this is the same enforce-then-delegate pattern used by all other modules in the codebase (`credstore`, `simple-user-settings`, etc.). Pushing AccessScope into every repo method would be a cross-cutting refactor across the entire platform, not scoped to this module.

modules/system/resource-group/resource-group/src/domain/validation.rs
```diff
@@ -0,0 +1,91 @@
...
+            "Type code must start with prefix '{RG_TYPE_PREFIX}', got: '{code}'"
+        )));
+        // @cpt-end:cpt-cf-resource-group-algo-type-mgmt-validate-type-input:p1:inst-val-input-2a
+    }
+    // @cpt-end:cpt-cf-resource-group-algo-type-mgmt-validate-type-input:p1:inst-val-input-2
+    // @cpt-begin:cpt-cf-resource-group-algo-type-mgmt-validate-type-input:p1:inst-val-input-3
+    if code.len() > 1024 {
+        return Err(DomainError::validation(
+            "Type code must not exceed 1024 characters",
+        ));
```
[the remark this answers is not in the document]
ffedoroff: Valid observation. In practice, type codes and group names are ASCII-only (enforced by the prefix/format validation above). Byte length == char count for our use case. No change needed.

modules/system/resource-group/resource-group/src/domain/error.rs
```diff
@@ -0,0 +1,209 @@
...
+            DomainError::GroupNotFound { id } => ResourceGroupError::not_found(id.to_string()),
+            DomainError::MembershipNotFound { key } => ResourceGroupError::not_found(key),
+            DomainError::TenantIncompatibility { message } => {
+                ResourceGroupError::tenant_incompatibility(message)
+            }
+            DomainError::AccessDenied { .. } => ResourceGroupError::internal(),
+            DomainError::Database { .. } | DomainError::InternalError => {
+                ResourceGroupError::internal()
+            }
```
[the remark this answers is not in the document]
ffedoroff: Not changing. Both PDP compile failures and denials produce 403 Forbidden — this matches the HTTP semantics (the server understood the request but refuses to authorize it). Splitting them into separate HTTP status codes would leak internal PDP implementation details to the client. If we ever need to distinguish them, we can add a sub-code in the error body without changing the HTTP status.

modules/system/resource-group/resource-group-sdk/src/error.rs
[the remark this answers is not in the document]
ffedoroff: No further changes needed. The SDK error layer is intentionally coarse — one `NotFound { code: String }` variant covers groups, types, and memberships. The domain layer has granular variants (`GroupNotFound { id: Uuid }`, `TypeNotFound { code }`, `MembershipNotFound { key }`) which all map to the same SDK `NotFound`. The `code` field holds whatever string identifier the caller used — a type path, a UUID string, or a composite key. This is the same pattern as other SDK error enums in the codebase (`credstore-sdk`, `types-registry-sdk`). Splitting into per-resource variants in the SDK would force every SDK consumer to handle N variants that all mean "404 — thing not found".

modules/system/resource-group/resource-group-sdk/src/odata/hierarchy.rs
```diff
@@ -0,0 +1,56 @@
...
+    /// Filter by GTS type path.
+    Type,
+}
+
+impl FilterField for HierarchyFilterField {
+    const FIELDS: &'static [Self] = &[Self::HierarchyDepth, Self::Type];
+
+    fn name(&self) -> &'static str {
+        match self {
+            Self::HierarchyDepth => "hierarchy/depth",
```
aviator5, line 24: I'm not sure that we have a convention for using "/" as a words separator in OData filters. Please check https://github.com/cyberfabric/DNA/blob/main/REST/QUERYING.md. I guess it should be hierarchy_depth, right?

modules/system/resource-group/resource-group-sdk/src/api.rs
```diff
@@ -0,0 +1,173 @@
...
+        id: Uuid,
+        request: PatchGroupRequest,
+    ) -> Result<ResourceGroup, ResourceGroupError>;
+
+    /// Delete a resource group. Fails if child groups or memberships exist.
+    async fn delete_group(&self, ctx: &SecurityContext, id: Uuid)
+    -> Result<(), ResourceGroupError>;
+
+    /// List group hierarchy with relative depth from a reference group.
+    async fn list_group_depth(
```
aviator5, line 116: [minor] Perhaps `get_group_descendants` is a bit clearer name.

modules/system/resource-group/resource-group-sdk/src/models.rs
```diff
@@ -0,0 +1,606 @@
...
+use std::fmt;
+
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+// -- GtsTypePath value object --
+
+// @cpt-begin:cpt-cf-resource-group-algo-sdk-foundation-validate-gts-type-path:p1:inst-gts-val-1
+/// Maximum length of a GTS type path.
+const GTS_TYPE_PATH_MAX_LEN: usize = 255;
```
aviator5: According to https://github.com/GlobalTypeSystem/gts-spec, max length GTS ID is 1024.

modules/system/resource-group/resource-group-sdk/src/models.rs
```diff
@@ -0,0 +1,606 @@
...
+    /// # Errors
+    /// Returns an error if the string is empty, exceeds 255 characters,
+    /// or does not match the GTS type path format.
+    pub fn new(raw: impl Into<String>) -> Result<Self, String> {
+        // @cpt-begin:cpt-cf-resource-group-algo-sdk-foundation-validate-gts-type-path:p1:inst-gts-val-2
+        let raw = raw.into();
+        let s = raw.trim().to_lowercase();
+        // @cpt-end:cpt-cf-resource-group-algo-sdk-foundation-validate-gts-type-path:p1:inst-gts-val-2
+
+        // @cpt-begin:cpt-cf-resource-group-algo-sdk-foundation-validate-gts-type-path:p1:inst-gts-val-3
```
aviator5, line 42: It's very inconvenient to review such code with eyes. I'm talking about all of these `@cpt-begin:`/`@cpt-end`.
@Artifizer, @ainetx let's do something with this

modules/system/resource-group/resource-group-sdk/src/models.rs
```diff
@@ -0,0 +1,606 @@
...
+    #[must_use]
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    /// Validate format: `gts.<segment>~(<segment>~)*`
+    /// where segment = `[a-z0-9_.]+`
+    #[allow(unknown_lints)]
+    #[allow(de0901_gts_string_pattern)]
+    fn matches_format(s: &str) -> bool {
```
aviator5: Please use https://github.com/globaltypesystem/gts-rust for parsing GTS expressions

modules/system/resource-group/resource-group/src/domain/validation.rs
```diff
@@ -0,0 +1,96 @@
+// @cpt-begin:cpt-cf-resource-group-dod-sdk-foundation-sdk-models:p1:inst-validation-full
+//! Shared domain validation utilities.
+
+use crate::domain::error::DomainError;
+
+/// GTS type path prefix required for resource group types.
+pub const RG_TYPE_PREFIX: &str = "gts.x.system.rg.type.v1~";
```
aviator5: `gts.x.system` or `gts.x.core` or `gts.cf.core`?

@Artifizer let's decide and define canonical prefix somewhere. I believe `docs/modkit_unified_system/14_gts.md` is a good candidate for this.

modules/system/resource-group/resource-group/src/domain/validation.rs
```diff
@@ -0,0 +1,96 @@
...
+///
+/// Returns `Ok(())` when:
+/// - `schema` is `None` (no schema = no validation, any metadata accepted)
+/// - `metadata` is `None` (nothing to validate)
+/// - `metadata` validates against the schema
+///
+/// # Errors
+///
+/// Returns [`DomainError`] when metadata violates the schema constraints.
+pub fn validate_metadata_against_schema(
```
aviator5: I believe https://github.com/globaltypesystem/gts-rust should be used for metadata validation.

modules/system/resource-group/resource-group-sdk/src/models.rs
ffedoroff: Fixed — changed `GTS_TYPE_PATH_MAX_LEN` from 255 to 1024 to match the [gts-spec](https://github.com/GlobalTypeSystem/gts-spec).

modules/system/resource-group/resource-group-sdk/src/models.rs
ffedoroff: Fixed — replaced the hand-rolled `matches_format()` parser with `gts::GtsID::new()` from the [gts-rust](https://github.com/globaltypesystem/gts-rust) crate (already in workspace as `gts = "0.8.4"`). Removed ~30 lines of manual segment validation.

modules/system/resource-group/resource-group/src/domain/validation.rs
ffedoroff: Fixed — changed `RG_TYPE_PREFIX` from `gts.x.system.rg.type.v1~` to `gts.x.core.rg.type.v1~` to match the prefix used by types-registry. Updated all references across SDK code, domain code, and 6 docs files (PRD, DESIGN, rust-traits, feature specs).

modules/system/resource-group/resource-group-sdk/src/models.rs
Artifizer: It's supposed you just ignore these lines, it's not for humans, it's for robots

modules/system/resource-group/resource-group/src/domain/validation.rs
Artifizer: Since now we have to use `gts.cf.core` as prefix for the core parts. The `gts.x` to be renamed everywhere, we can have a dylint for that as well

modules/system/resource-group/resource-group-sdk/src/api.rs
ffedoroff: depth could be negative and represent ascendants not only descendants, that is why I keep depth word here

modules/system/resource-group/resource-group-sdk/src/odata/hierarchy.rs
ffedoroff: Per [OData 4.0 §5.1.1.7 "Path Expressions"](https://docs.oasis-open.org/odata/odata/v4.0/errata02/os/complete/part2-url-conventions/odata-v4.0-errata02-os-part2-url-conventions-complete.html):

> *"Properties of complex properties can be used via the same syntax as in resource paths, i.e. by specifying the name of a complex property, followed by a forward slash (/) and the name of a property of the complex property"*

In our case `hierarchy` is a complex type in the response:

```json
{
  "id": "...",
  "name": "...",
  "hierarchy": {
    "depth": 2,
    "parent_id": "..."
  }
}
```

So `hierarchy/depth` and `hierarchy/parent_id` are valid OData property path expressions for filtering on nested fields of a complex type.

That said, [DNA/QUERYING.md](https://github.com/cyberfabric/DNA/blob/main/REST/QUERYING.md) doesn't mention complex type navigation — all examples use flat `snake_case` fields (`created_at`, `id`, `status`). So the project convention is currently undefined for this case.

Two options:
1. **Keep `hierarchy/depth`** — follows OData standard, matches the JSON response structure
2. **Rename to `hierarchy_depth`** — follows the flat naming pattern from DNA, simpler but disconnected from the response shape

I'd suggest we decide on a convention for complex type filter fields in DNA and apply it consistently. What do you think?

modules/system/resource-group/resource-group/src/domain/validation.rs
ffedoroff: Updated to `gts.cf.core` — replaced all 101 occurrences across SDK code, domain code, docs (PRD, DESIGN, rust-traits, feature specs). No `gts.x.core` or `gts.x.system` left in the resource-group module.

modules/system/resource-group/resource-group/src/domain/validation.rs
ffedoroff: Agreed. In this push:

- **GTS ID validation** (`GtsTypePath::new`) now uses `gts::GtsID::new()` from the gts crate directly. All chained type path segments use 5-token format (`vendor.package.namespace.type.vMAJOR`) which `GtsID::new()` validates natively. No manual parser needed.
- **Metadata validation** (`validate_metadata_via_gts`) resolves the type's schema through `TypesRegistryClient` (same pattern as credstore module) for full GTS-level validation (`allOf` composition, `$ref` resolution, `x-gts-traits`). If the type is not yet registered in GTS (e.g. during initial seeding), validation is skipped gracefully.
- `gts = { workspace = true }` added to resource-group-sdk deps.

modules/system/resource-group/resource-group-sdk/src/odata/hierarchy.rs
nonameffh: From my perspective "hierarchy/depth" can be kept if it's per OData specification, but such filters should be accepted only when white-listed, otherwise we could easily introduce a possible vulnerability to DDoS attacks and performance degradation

ffedoroff: so after conversation in other chats, we decided to keep option 1 `hierarchy/depth` - follows OData standard

modules/system/resource-group/resource-group-sdk/src/models.rs
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    /// Validate format: `gts.<segment>~(<segment>~)*`
+    /// where segment = `[a-z0-9_.]+`
+    ///
+    /// Uses the `gts` crate to validate the first (schema) segment, then
+    /// validates remaining chained segments with the same character rules.
+    #[allow(unknown_lints)]
+    #[allow(de0901_gts_string_pattern)]
+    fn is_valid_gts_path(s: &str) -> bool {
```
aviator5: Why do we need to implement gts validation manually?
I already had the similar question in https://github.com/cyberfabric/cyberfabric-core/pull/1492#discussion_r3084724612

Do we really need `GtsTypePath` at all?

modules/system/resource-group/resource-group-sdk/src/api.rs
aviator5: hm, it make API not obvious.. For me it's much better to have a couple of methods `get_group_descendants` and `get_group_ancestors`. I think they will be used in completely different scenarios. But ok, it's up to you

modules/system/resource-group/resource-group-sdk/src/odata/memberships.rs
```diff
@@ -0,0 +1,43 @@
...
+            Self::GroupId => "group_id",
+            Self::ResourceType => "resource_type",
+            Self::ResourceId => "resource_id",
+        }
+    }
+
+    fn kind(&self) -> FieldKind {
+        match self {
+            Self::GroupId => FieldKind::Uuid,
+            Self::ResourceType => FieldKind::I64,
```
MikeFalcon77: **CRITICAL**

`MembershipFilterField::ResourceType` uses `FieldKind::I64` but `resource_type` is a GTS type path string in the public API.

DESIGN §2.2 (Surrogate IDs Are Internal Only): "SMALLINT surrogate IDs MUST NOT appear in any API response, SDK type, REST contract, or OpenAPI schema. All external interfaces use GTS type paths (strings) exclusively." OData filter consumers will pass string type paths like `gts.z.system.idp.user.v1~`, which will fail to parse as `I64`.

Change to `FieldKind::String`. The persistence layer should resolve string paths to SMALLINT IDs after OData parsing, same as `GroupFilterField::Type` already does.

modules/system/resource-group/resource-group-sdk/src/odata/memberships_tests.rs
```diff
@@ -0,0 +1,17 @@
+// Created: 2026-04-16 by Constructor Tech
+use super::*;
+
+// TC-ODATA-05: MembershipFilterField names + kinds
+#[test]
+fn membership_filter_field_names_and_kinds() {
+    assert_eq!(MembershipFilterField::GroupId.name(), "group_id");
+    assert_eq!(MembershipFilterField::GroupId.kind(), FieldKind::Uuid);
```
MikeFalcon77, line 8: **CRITICAL**

This test asserts the wrong `FieldKind` for `ResourceType`. It validates the bug instead of catching it.

`resource_type` is a GTS type path (string) in the public API, so the assertion should be `FieldKind::String`, not `FieldKind::I64`. Update to: `assert_eq!(MembershipFilterField::ResourceType.kind(), FieldKind::String);`

modules/system/resource-group/resource-group-sdk/src/api.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// Patch a resource group (partial update via PATCH).
+    async fn patch_group(
+        &self,
+        ctx: &SecurityContext,
+        id: Uuid,
+        request: PatchGroupRequest,
+    ) -> Result<ResourceGroup, ResourceGroupError>;
+
+    /// Delete a resource group. Fails if child groups or memberships exist.
+    async fn delete_group(&self, ctx: &SecurityContext, id: Uuid)
```
MikeFalcon77: **HIGH**

`delete_group` is missing the `force: bool` parameter.

DESIGN §3.3 API Contracts specifies: `delete_group` → "delete group (optional `force`)". DESIGN §3.9 Error Mapping specifies `ConflictActiveReferences` for non-force delete when children/memberships exist. The reference traits document (`rust-traits.md`) includes `force: bool` in the signature. PRD §5.2 also defines force-delete as a functional requirement (`cpt-cf-resource-group-fr-force-delete`).

Add `force: bool` parameter: `async fn delete_group(&self, ctx: &SecurityContext, id: Uuid, force: bool) -> Result<(), ResourceGroupError>;`

modules/system/resource-group/resource-group/src/domain/repo.rs
```diff
@@ -0,0 +1,288 @@
...
+    ) -> Result<ResourceGroupType, DomainError>;
+
+    async fn resolve_id<C: DBRunner>(&self, db: &C, code: &str)
+    -> Result<Option<i16>, DomainError>;
+
+    async fn insert<C: DBRunner>(
+        &self,
+        db: &C,
+        schema_id: &str,
+        can_be_root: bool,
```
MikeFalcon77: **HIGH**

`TypeRepositoryTrait::insert` accepts `can_be_root: bool` but the `gts_type` entity has no `can_be_root` column.

DESIGN §3.7 (Table: `gts_type`) explicitly states: "`can_be_root` is resolved at runtime from `x-gts-traits` in the registered GTS schema (not a DB column)". The repo trait cannot persist this value. Same issue in `update_type` at line 212.

Remove `can_be_root: bool` from both `insert` and `update_type` signatures. Resolve `can_be_root` in the domain service layer by querying the types registry (`TypesRegistryClient`) — analogous to how `validate_metadata_via_gts` already calls the types registry.

modules/system/resource-group/resource-group/src/domain/repo.rs
```diff
@@ -0,0 +1,288 @@
...
+use modkit_db::secure::DBRunner;
+use modkit_odata::{ODataQuery, Page};
+use modkit_security::AccessScope;
+use resource_group_sdk::models::{
+    ResourceGroup, ResourceGroupMembership, ResourceGroupType, ResourceGroupWithDepth,
+};
+use uuid::Uuid;
+
+use crate::domain::error::DomainError;
+use crate::infra::storage::entity::{
```
MikeFalcon77, line 13: **MEDIUM**

Domain repository traits import infrastructure entity types (`gts_type::Model`, `rg_entity::Model`, `membership_entity::Model`). Methods like `find_model_by_id` return `rg_entity::Model` and `load_full_type` accepts `&gts_type::Model`.

This violates DDD layering: domain layer depends on infrastructure. The suppressed lint `de0301_no_infra_in_domain` confirms this is a known violation. For a contracts-first PR this may be acceptable, but domain repo traits should use domain types (or SDK types where appropriate), with entity↔domain conversion in the infra layer.

Consider defining domain-level result types for repo methods (e.g., domain `TypeRecord` instead of `gts_type::Model`), or move the entity-returning methods to an infra-level trait that the service layer uses directly.

modules/system/resource-group/resource-group/src/domain/error.rs
```diff
@@ -0,0 +1,213 @@
...
+                ResourceGroupError::conflict_active_references(message)
+            }
+            DomainError::Conflict { message } => ResourceGroupError::conflict(message),
+            DomainError::GroupNotFound { id } => ResourceGroupError::not_found(id.to_string()),
+            DomainError::MembershipNotFound { key } => ResourceGroupError::not_found(key),
+            DomainError::TenantIncompatibility { message } => {
+                ResourceGroupError::tenant_incompatibility(message)
+            }
+            DomainError::AccessDenied { .. } => ResourceGroupError::access_denied(),
+            DomainError::Database { .. } | DomainError::InternalError => {
```
MikeFalcon77, line 178: **MEDIUM**

`DomainError::Database` always maps to `ResourceGroupError::Internal`. DESIGN §3.9 Error Mapping specifies `ServiceUnavailable` for "infra timeout/unavailable".

`DomainError` has no variant that distinguishes transient infrastructure failures (connection timeout, DB unavailable) from permanent internal errors. The `ResourceGroupError::ServiceUnavailable` variant exists but is unreachable from domain error conversion.

Consider splitting `DomainError::Database` into infrastructure-failure categories (e.g., `DatabaseUnavailable` → `ServiceUnavailable`, `DatabaseError` → `Internal`), or add a `DomainError::ServiceUnavailable` variant.

modules/system/resource-group/resource-group-sdk/src/models_tests.rs
```diff
@@ -0,0 +1,226 @@
...
+    assert!(
+        json.get("metadata").is_none(),
+        "metadata should be absent when None, got: {json}"
+    );
+}
+
+// -- QueryProfile default -- TC-SDK-17
+
+#[test]
+fn query_profile_default_values() {
```
MikeFalcon77: **MEDIUM**

`query_profile_default_values` test has an empty body with only a comment. This is a no-op test that inflates test count without verifying any behavior.

Either implement the test if `QueryProfile` is accessible from SDK tests (e.g., via re-export), or remove it entirely. A placeholder test that passes unconditionally is worse than no test — it creates false confidence in coverage.

modules/system/resource-group/resource-group-sdk/src/models.rs
```diff
@@ -0,0 +1,387 @@
...
+    pub metadata: Option<serde_json::Value>,
+}
+
+/// Serde helper for `Option<Option<T>>` that distinguishes absent, null, and present values.
+///
+/// - `None` (absent): field is skipped (use `skip_serializing_if = "Option::is_none"`)
+/// - `Some(None)` (null): serializes as `null`
+/// - `Some(Some(v))` (present): serializes as the inner value
+#[allow(clippy::option_option, clippy::missing_errors_doc, clippy::ref_option)]
+pub mod option_option {
```
MikeFalcon77, line 275: **MEDIUM**

The `option_option` serde module implements tri-state semantics (absent / null / value) for `PatchGroupRequest` but has zero test coverage.

Tri-state `Option<Option<T>>` serialization is subtle and error-prone. Without tests, regressions in PATCH semantics (e.g., `null` being interpreted as absent, or absent fields being serialized as `null`) will not be caught.

Add tests covering: (1) field absent → `None` after deserialization, (2) field `null` → `Some(None)`, (3) field with value → `Some(Some(v))`, (4) round-trip serialization preserves semantics, (5) `skip_serializing_if` correctly omits `None` fields.

modules/system/resource-group/resource-group-sdk/src/models.rs
aviator5: Sorry, but I don't understand why we are going to introduce a new "GTS path" which is not a valid GTS id.

Additionally, comment says "Uses the `gts` crate to validate the first (schema) segment" that is wrong.

modules/system/resource-group/resource-group-sdk/src/api.rs
ffedoroff: We considered splitting but there's one common scenario where a single call returning both directions is useful: building a tree view with breadcrumb navigation.

For example, when a user opens "Group A", the UI needs both ancestors (breadcrumb) and subgroups in one request:

```
          Tenant Corp          ← depth -2 (ancestor)
             │
        Organization HQ        ← depth -1 (ancestor)
             │
        ► Group A ◄            ← depth  0 (current)
          ├── Subgroup Alpha   ← depth +1
          ├── Subgroup Beta    ← depth +1
          │     └── Team X     ← depth +2
          └── Subgroup Gamma   ← depth +1
```

With `list_group_depth` and `$filter=hierarchy/depth ge -10 and hierarchy/depth le 10`, a single API call returns the full context — ancestor chain for breadcrumb and subgroups for the tree. Two separate methods would require two round-trips for this common screen.

modules/system/resource-group/resource-group/src/domain/repo.rs
ffedoroff: Known pragmatic trade-off — all modules in this codebase follow the same pattern where domain repository traits reference entity types from the infrastructure layer (`gts_type::Model`, `rg_entity::Model`). Refactoring to pure domain models would be a cross-cutting change across the entire platform. The current approach is consistent with credstore, nodes-registry, and other modules.

modules/system/resource-group/resource-group/src/domain/error.rs
ffedoroff: `Database` errors map to `Internal` because DB failures can contain sensitive information (table names, SQL fragments) that should not leak to clients. The DESIGN §3.9 `ServiceUnavailable` mapping refers to transient infrastructure failures (connection timeouts, pool exhaustion) which are a subset of DB errors. Currently all DB errors are treated as internal — distinguishing transient vs permanent would require inspecting `sea_orm::DbErr` variants, which we can add when there's an operational need for client-side retry logic based on the error type.

modules/system/resource-group/resource-group-sdk/src/models_tests.rs
ffedoroff: Fixed — removed the empty `query_profile_default_values` test. `QueryProfile` lives in the domain crate, not SDK — it's tested in `domain_unit_test.rs`.

modules/system/resource-group/resource-group-sdk/src/models.rs
ffedoroff: The `option_option` serde module is exercised through `PatchGroupRequest` integration tests in `api_rest_test.rs` — specifically `rest_patch_group_updates_name_only`, `rest_patch_group_clears_metadata_with_null`, and `rest_patch_group_preserves_fields_not_in_request`. These test the full serde round-trip (JSON → PatchGroupRequest → service → response). Dedicated unit tests for the module itself would duplicate this coverage.

modules/system/resource-group/resource-group-sdk/src/models.rs
ffedoroff: Fixed in the latest push — `is_valid_gts_path` is removed entirely. `GtsTypePath::new()` now calls `gts::GtsID::new()` directly:

```rust
if gts::GtsID::new(&s).is_err() {
    return Err("Invalid GTS type path format".to_owned());
}
```

All chained segments updated to 5-token format (`vendor.package.namespace.type.vMAJOR`) so they pass `GtsID` validation natively. No custom parser, no deviation from GTS spec.

modules/system/resource-group/resource-group-sdk/src/odata/memberships.rs
ffedoroff: Fixed — changed `MembershipFilterField::ResourceType` from `FieldKind::I64` to `FieldKind::String`. Test assertion updated accordingly. Good catch.

modules/system/resource-group/resource-group-sdk/src/odata/memberships_tests.rs
ffedoroff: Fixed — test assertion updated from `FieldKind::I64` to `FieldKind::String`.

modules/system/resource-group/resource-group-sdk/src/api.rs
ffedoroff: Fixed — added `force: bool` parameter to the SDK trait:

```rust
async fn delete_group(&self, ctx: &SecurityContext, id: Uuid, force: bool)
    -> Result<(), ResourceGroupError>;
```

When `force` is false, fails with `ConflictActiveReferences` if child groups or memberships exist. When `force` is true, recursively deletes the entire subtree. The domain service `GroupService::delete_group` and REST handler `?force=true` already support this — the SDK trait was the missing piece.

modules/system/resource-group/resource-group/src/domain/repo.rs
ffedoroff: Fixed — removed `can_be_root: bool` from both `TypeRepositoryTrait::insert` and `TypeRepositoryTrait::update_type`. Per DESIGN §3.7, `can_be_root` is resolved at runtime from `x-gts-traits` via `TypesRegistryClient`, not persisted as a DB column. The repo layer now only stores `schema_id` and `metadata_schema`. Updated docs to reflect this.

modules/system/resource-group/resource-group-sdk/src/api.rs
ffedoroff: after more conversations decided to split to get_group_descendants and get_group_ancestors

Artifizer: > depth could be negative and represent ascendants not only descendants, that is why I keep depth word here

I think negative depth value could be misleading and error prone in contracts in general especially if it's propagated further without an explicit check. It is probably not critical right now, but it looks fragile and may lead to incorrect behavior with future changes. e.g. the upstream/clients can get a negative array index, etc

aviator5: @ffedoroff sorry, but there is still `list_group_depth` instead of `get_group_descendants`/`get_group_ancestors`

modules/system/resource-group/resource-group-sdk/src/error.rs
```diff
@@ -0,0 +1,151 @@
...
+    #[error("Cycle detected: {message}")]
+    CycleDetected { message: String },
+
+    /// A configured limit (depth, width, etc.) would be exceeded.
+    #[error("Limit violation: {message}")]
+    LimitViolation { message: String },
+
+    /// Tenant scope incompatibility.
+    #[error("Tenant incompatibility: {message}")]
+    TenantIncompatibility { message: String },
```
aviator5, line 58: Could you please clarify when `TenantIncompatibility` can be returned from the RG service?

modules/system/resource-group/resource-group-sdk/src/models.rs
```diff
@@ -0,0 +1,352 @@
...
+///
+/// Group responses do NOT include `created_at`/`updated_at` (per DESIGN).
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ResourceGroup {
+    /// Group identifier.
+    pub id: Uuid,
+    /// GTS chained type path.
+    #[serde(rename = "type")]
+    pub type_path: String,
```
aviator5: Suggestion: `type_path` -> `type_code`?
To be consistent with `ResourceGroupType.code`.

modules/system/resource-group/resource-group-sdk/src/api.rs
```diff
@@ -0,0 +1,181 @@
...
+    ) -> Result<Page<ResourceGroupWithDepth>, ResourceGroupError>;
+}
+
+// @cpt-flow:cpt-cf-resource-group-flow-integration-auth-plugin-routing:p1
+/// Extended read trait for vendor-specific plugin gateway routing.
+///
+/// Extends `ResourceGroupReadHierarchy` with membership listing for
+/// plugin consumers that need both hierarchy and membership data.
+#[async_trait]
+pub trait ResourceGroupReadPluginClient: ResourceGroupReadHierarchy {
```
aviator5: Do we really need two traits for the AuthZ case? Why a single trait with 3 methods (`list_memberships`, `get_group_acenstors`, `get_group_descendants`) is not enough?

modules/system/resource-group/resource-group-sdk/src/error.rs
ffedoroff: `TenantIncompatibility` is raised in exactly one place — `MembershipService::add_membership` — when a resource is already a member of a group in a different tenant than the target group.

## Invariant

Each resource, identified by the pair `(resource_type, resource_id)`, must belong to groups in **one tenant only**. Adding it to a group that lives in a different tenant would create a cross-tenant link and is rejected at the service layer.

## Where it is produced

`modules/system/resource-group/resource-group/src/domain/membership_service.rs` (visible in the follow-up backend PR, not in this PR which ships contracts only):

```rust
// existing_tenants — distinct tenant_ids across memberships of this (resource_type, resource_id)
if !existing_tenants.is_empty() && !existing_tenants.contains(&group_model.tenant_id) {
    return Err(DomainError::tenant_incompatibility(format!(
        "Resource ({resource_type}, {resource_id}) is already linked in tenant {:?}, \
         cannot add to tenant {}",
        existing_tenants, group_model.tenant_id
    )));
}
```

The check is cheap: one `SELECT DISTINCT tenant_id` over existing memberships of the resource, short-circuited when the resource has no memberships yet.

## Example

1. `add_membership(group = T1.Dept1, resource = (user, alice))` → stored, `memberships.tenant_id = T1`.
2. Later: `add_membership(group = T2.Branch1, resource = (user, alice))` →
   `existing_tenants = {T1}`, `group.tenant_id = T2` → rejected with
   `TenantIncompatibility { message: "Resource (user, alice) is already linked in tenant {T1}, cannot add to tenant T2" }`.

Adding `alice` to **another** group within T1 (e.g. `T1.Dept2`) succeeds because `{T1}.contains(T1)` holds.

## Why the invariant is enforced at service layer

- **Data isolation.** A single resource belongs to one tenant by contract. Without this check, the membership table could reference the same resource from two tenants at once, which breaks the fundamental tenant-scoping guarantee the whole platform relies on.
- **AuthZ predicate simplicity.** The `In(owner_tenant_id, …)` predicate emitted by `rg-authz-plugin`/`tr-authz-plugin` assumes each row belongs to exactly one tenant. Cross-tenant rows would silently leak across scopes.
- **Barrier / self-managed semantics.** A self-managed tenant must be a strict isolation boundary. If a resource inside a self-managed tenant were also linked from a parent tenant, the parent would effectively see inside the barrier via that shared resource.

## Where it is consumed (not produced)

`seeding::seed_memberships` treats it as a graceful skip rather than a fatal error — seeders can contain stale records across environments, so the service returns the error but the seeder logs a warning and continues:

```rust
Err(DomainError::TenantIncompatibility { .. }) => {
    tracing::warn!(…, "Skipping membership seed: tenant incompatibility");
    result.skipped += 1;
}
```

## Why the variant ships in the SDK crate already

Consumers writing against `ResourceGroupClient::add_membership` need an exhaustive set of error variants from day one. Adding a new variant later would be a breaking change for downstream crates matching on `ResourceGroupError`. Shipping it now keeps PR 1/4 (contracts) forward-compatible with the backend PR that actually raises it.

aviator5: Got it, thanks for the explanation, suggesting adding few words in comments, otherwise it's not obvious for now.

modules/system/resource-group/resource-group-sdk/src/api.rs
ffedoroff: Agreed — `ResourceGroupReadPluginClient` had no consumers, removed it. The integration read surface is now a single 2-method `ResourceGroupReadHierarchy` (`get_group_descendants` + `get_group_ancestors`). `list_memberships` stays on `ResourceGroupClient` where it actually belongs.

modules/system/resource-group/resource-group-sdk/src/models.rs
ffedoroff: Fixed — renamed to `code` in all four group structs (`ResourceGroup`, `ResourceGroupWithDepth`, `CreateGroupRequest`, `UpdateGroupRequest`) for full symmetry with `ResourceGroupType.code`. JSON wire key stays `"type"` via `#[serde(rename = "type")]`, so REST/OpenAPI are unaffected.

modules/system/resource-group/resource-group-sdk/src/error.rs
ffedoroff: Done — expanded the docstring on both `ResourceGroupError::TenantIncompatibility` (SDK) and `DomainError::TenantIncompatibility` (domain) to make the raise condition explicit: cross-tenant `add_membership` when the resource already has memberships in a different tenant.

aviator5 (approved):

Artifizer (approved):

PR #1530, opened 2026-04-16
feat(oagw): implement rate limiting with hierarchical budget allocation

asmith987:
Wire token bucket and sliding window rate limiters into the proxy data plane with scope-aware keys (global/tenant/user/ip/route), x-ratelimit-* and retry-after response headers, cost-based deduction, and hierarchical budget allocation. Add e2e tests for budget allocation across hierarchy.

## Summary by CodeRabbit

* **New Features**
  * Hierarchical rate-limit budgets (Allocated/Shared/Unlimited), sliding-window algorithm, weighted costs, per-scope enforcement, and configurable X-RateLimit-* header emission.

* **Bug Fixes**
  * Safer, prefix-based cleanup of rate-limit keys on resource changes; more complete and consistent 429 responses (Retry-After plus optional X-RateLimit-* headers).

* **Documentation**
  * Updated rate-limiting key format and cleanup guidance.

* **Tests**
  * Expanded unit/integration/e2e coverage and added e2e fixtures/tokens for hierarchy, scope, budget, header, and error scenarios.

### DIFF
--- a/modules/system/oagw/oagw/src/api/rest/handlers/route.rs
+++ b/modules/system/oagw/oagw/src/api/rest/handlers/route.rs
@@ -105,11 +105,22 @@ pub async fn update_route(
 ) -> Result<impl IntoResponse, Problem> {
     let instance = format!("/oagw/v1/routes/{id}");
     let uuid = parse_gts_id(&id, gts::ROUTE_SCHEMA, &instance)?;
+    // Snapshot old rate_limit before update so we can detect changes and
+    // clean up stale rate-limit keys (avoids accumulating orphaned buckets).
+    let old_rate_limit = state
+        .cp
+        .get_route(&ctx, uuid)
+        .await
+        .map_err(|e| domain_error_to_problem(e, &instance))?
+        .rate_limit;
     let route = state
         .cp
         .update_route(&ctx, uuid, req.into())
         .await
         .map_err(|e| domain_error_to_problem(e, &instance))?;
+    if route.rate_limit != old_rate_limit {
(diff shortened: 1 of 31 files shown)


modules/system/oagw/oagw/src/domain/services/management.rs
```diff
@@ -478,6 +499,115 @@ impl ControlPlaneServiceImpl {
...
+
+        let budget_total = match ancestor_budget.total {
+            Some(t) => t,
+            None => return Ok(()), // No total specified — nothing to enforce.
+        };
+        let ratio = ancestor_budget.overcommit_ratio.unwrap_or(1.0);
+
+        // Get all upstreams with this alias, then filter to the ancestor's
+        // tenant tree to avoid counting upstreams from unrelated trees.
+        let all_siblings = self.upstreams.list_by_alias(alias).await?;
```
mattgarmon: I think better to `get_descentents` first, and filter this query by those returned tenants.

Ideally, we would avoid the `get_descendents` all together, but from the tests and docs it seems that "all child allocations"  need to be considered.

modules/system/oagw/oagw/src/api/rest/handlers/route.rs
mattgarmon, line 100: We should be calling `remove_rate_limit_keys_for_route` when the rate-limit key changes, to avoid accumulating state keys. 

This also applies to `update_upstream`.

We could also do it unconditionally, but this could allow rate-limit bypasses when unrelated properties are updated.

mattgarmon (approved):

Artifizer (approved):

PR #1532, opened 2026-04-17
fix(bootstrap): unify crypto provider init in main() for both FIPS and non-FIPS modes

MikeFalcon77:
After the streamline refactor (#1331) moved FIPS crypto init from main() into init_procedure(), the provider was installed too late — after config loading. For non-FIPS mode, no provider was installed at all, causing a rustls panic when both ring and aws-lc-rs are compiled in (#1502 patched non-FIPS only).

Consolidate both cases into a single init_crypto_provider() function in modkit::bootstrap, called as the first line of main(). This ensures the crypto provider is always installed before any TLS, DB, or JWT operation regardless of the feature flag.

## Summary by CodeRabbit

* **Chores**
  * Refactored internal cryptographic provider initialization to improve code organization and maintainability.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,0 +1,5 @@
+# Maintainers
+
+artifizer@gmail.com
+vzgrab@gmail.com
+aa@acronis.com
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,11 +1,5 @@
 # Security Policy
 
-## Supported Versions
-
-| Version | Supported          |
-| ------- | ------------------ |
-| 0.1.x   | :white_check_mark: |
-
 ## Reporting Vulnerabilities
 
 **Do NOT report security vulnerabilities through public GitHub issues.**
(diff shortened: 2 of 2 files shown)


apps/hyperspot-server/src/main.rs
```diff
@@ -64,15 +64,9 @@ enum Commands {
...
-    // Install aws-lc-rs as the default rustls CryptoProvider before any TLS
-    // config is constructed. Required because both `ring` and `aws-lc-rs`
-    // providers are compiled in (ring via aliri/pingora-rustls), and rustls
-    // 0.23 panics if it cannot auto-detect a single provider.
-    // Skip in FIPS mode — init_procedure() installs the FIPS provider instead.
-    #[cfg(not(feature = "fips"))]
-    if let Err(_e) = rustls::crypto::aws_lc_rs::default_provider().install_default() {
-        // Another provider was already installed — safe to continue.
-    }
+    // Install the crypto provider before anything else.
```
MikeFalcon77: test passed

PR #1570, opened 2026-04-20
feat(tracing): deepen instrumentation across host, handlers, auth, DB, and retries

### LINKED ISSUE #817: Tracing instrumentation is shallow
## TL;DR

Observability is partially broken: trace IDs surfaced to clients are incomplete, there is no span hierarchy (so no timing breakdown), and entire layers — auth, database, module handlers, startup phases — produce no spans at all. The OTEL provider is also never shut down, silently dropping the final batch of spans on exit.

---

Distributed traces for a typical request produce a single `http_request` span with log events inside it. There is no span hierarchy, so there is no way to see "auth took 5ms, handler took 20ms, DB query took 12ms." Multiple layers — auth, database, module handlers — emit no spans at all, and the trace ID surfaced to clients in `Problem` responses doesn't match the real OTEL trace ID.

1. **Trace ID extraction is inconsistent and broken.** Three separate mechanisms produce three different values: `trace_layer.rs` emits `"Id(42)"` (Debug format of a subscriber-local ID), `error_layer.rs` emits `"42"` (raw u64), and the api-gateway `TraceLayer` records the real W3C trace ID. `Problem` responses use the first two, so clients see useless IDs that can't be correlated with Jaeger/Tempo. The `traceparent` fallback in `error_layer.rs` returns the full W3C header string rather than just the trace ID portion.

2. **Span hierarchy is flat.** Across the codebase there are ~8 uses of `#[tracing::instrument]` (all in `lifecycle.rs`) and ~3 manual span creation sites (`rpc_retry.rs`, `layers/otel.rs`, api-gateway `TraceLayer`), vs. 200+ event macros (`info!`, `warn!`, `error!`, etc.) spread across ~40 files. Nearly all tracing is events, not timed spans, so no timing breakdown is possible.

3. **Module handlers are un-instrumented.** Of 6 modules with REST handlers, only `file-parser` uses `#[instrument]`. The other five (`simple-user-settings`, `nodes-registry`, `types-registry`, `module-orchestrator`, `oagw`) create no handler spans, no entry logging, and record no fields.

4. **Database layer is invisible.** `SecureConn` — the chokepoint for all tenant-scoped DB access — has no tracing on `find_*`, `insert_*`, `update_*`, `delete_*`. Connections are built from raw sqlx pools bypassing SeaORM's `ConnectOptions::sqlx_logging()`. Transaction boundaries are not traced. It's impossible to attribute query time to handlers, see transaction scope, or detect N+1 patterns.

5. **Auth middleware is silent.** JWT validation, JWKS key refresh, token claims extraction, and permission checks create no spans. Auth success is entirely silent; failures appear only as debug-level events. A JWKS refresh adding 500ms latency is invisible in traces.

6. **HostRuntime phases have no spans.** `run_init_phase`, `run_db_phase`, `run_rest_phase`, `run_start_phase`, and the other phase methods log `info!` events at entry but don't create spans, so phase duration during startup is invisible.

7. **OTEL provider is never shut down.** The `SdkTracerProvider` is not stored after creation, so `shutdown()` is never called. The final batch of in-flight spans is silently dropped on process exit.

8. **HTTP client retries are events, not spans.** Retry attempts in `modkit-http` are logged as `debug!` events inside the `outgoing_http` span. A retry sequence of 3 attempts has no individual timing — unlike the gRPC retry path, which already uses per-attempt `debug_span!`.

## Suggestions

The fixes range from one-line utility replacements (trace ID) to systematic span additions across layers. Ordered by observability payoff:

- **Unify trace ID extraction behind a single OTEL-aware utility.** Use `tracing_opentelemetry::OpenTelemetrySpanExt` to get the real trace ID from the current span context, falling back to request headers. Replace all `Span::current().id()` call sites (`trace_layer.rs`, `error_layer.rs`, `problem.rs`, module `error.rs` files).
- **Store the `SdkTracerProvider` and call `shutdown()` on process exit.** A `OnceLock<SdkTracerProvider>` and a `shutdown_tracing()` call from the bootstrap path is sufficient.
- **Add `#[instrument]` to HostRuntime phase methods and per-module init loops.** Each phase gets a timed span; each module within a phase gets a child span. This gives startup visibility immediately.
- **Add a handler-span middleware using Axum's `MatchedPath`.** A single middleware creating an `info_span!("handler", "http.route" = %path)` and calling `next.run(request).instrument(span).await` gives automatic handler coverage without touching every module. Handlers that need more granularity can layer per-handler `#[instrument]` on top.
- **Wrap auth middleware in a span with `auth.method`, `auth.principal`, `auth.result` fields.** Record-later pattern for fields not known at span creation.
- **Add `#[instrument]` to `SecureConn` DB operations and transaction boundaries.** Use OTEL semantic conventions (`db.system`, `db.operation`). This attributes query time to the handler span above.
- **Convert HTTP client retry attempts from events to child spans.** Mirrors the existing gRPC retry pattern in `rpc_retry.rs`.
- **Wire in `tracing-error::ErrorLayer`** (already a workspace dep, never installed). Enables `SpanTrace::capture()` on error types for rich context on errors that bubble across layers.
- **Separate OTEL filter from console filter.** OTEL should export INFO+ only to avoid sending DEBUG/TRACE spans to the collector, while console keeps its own independent filter.
- **Standardize field naming across all spans/events.** Adopt OTEL semantic conventions (`error` not `err`, `http.method`, `db.system`, `auth.method`, etc.) and document the convention.

lizreu:
- Replace three broken `Span::current().id()`-based trace-ID extractors (plus five per-module copies) with `modkit::telemetry::current_trace_id()`, which reads the real W3C trace ID via `OpenTelemetrySpanExt`. Removes the `Problem::into_response` auto-enrich that manufactured fake trace IDs; callers set `trace_id` explicitly or via `WithTraceContext`.
- Store `SdkTracerProvider` and `SdkMeterProvider` in `OnceLock` so `shutdown_tracing` / `shutdown_metrics` actually force-flush and shut down. Previously documented no-ops — the final batch of spans was dropped on every process exit.
- Install `tracing_error::ErrorLayer` (already a workspace dep, never wired in).
- Span-instrument `HostRuntime::run_*_phase` methods (`phase="init"/"db"/...` + per-module child spans), the Axum handler layer (new `handler_span_middleware` populating `http.route` from `MatchedPath`), `authn_middleware` (`auth` span with late-bound `auth.method` / `auth.principal` / `auth.result`), `SecureConn`'s async DB ops and transaction boundaries (`db.system`, `db.operation`, `entity`), and `modkit-http`'s retry loop (per-attempt `http.retry` spans, parity with the gRPC retry path).
- Document the OTEL semantic-convention field naming in `docs/TRACING_SETUP.md`. Mass rename of existing fields is not done — log pipelines depend on the legacy names.

Note: `Problem.trace_id` is now `null` when the caller didn't set one, rather than the fake `"42"` the old auto-enrich produced. All in-tree callers set it explicitly, so in-tree behavior is unchanged; worth a line in release notes for external consumers.

Closes #817.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Handler-level HTTP request tracing (route, method, status) and new telemetry APIs for canonical trace-id extraction
  * Database spans for inserts/updates/deletes and transaction boundaries
  * Per-attempt HTTP retry spans and enhanced auth tracing (method/result/principal)
  * Improved tracer/metrics init and conditional force-flush/shutdown; tracing-error layer added

* **Bug Fixes**
  * Trace ID extraction/propagation aligned with OpenTelemetry; error responses no longer auto-enrich trace IDs

* **Documentation**
  * Full rewrite of tracing setup: OpenTelemetry wiring, YAML schema, span reference, and troubleshooting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1417,6 +1417,7 @@ dependencies = [
  "tower",
  "tracing",
  "tracing-appender",
+ "tracing-error",
  "tracing-log",
  "tracing-opentelemetry",
  "tracing-subscriber",
@@ -1542,7 +1543,6 @@ dependencies = [
  "http",
  "serde",
  "serde_json",
- "tracing",
  "utoipa",
 ]
 
@@ -8335,6 +8335,16 @@ checksum = "db97caf9d906fbde555dd62fa95ddba9eecfd14cb388e4f491a66d74cd5fb79a"
 dependencies = [
  "once_cell",
  "valuable",
+]
+
+[[package]]
+name = "tracing-error"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b1581020d7a273442f5b45074a6a57d5757ad0a47dac0e9f0bd57b81936f3db"
+dependencies = [
+ "tracing",
+ "tracing-subscriber",
 ]
 
 [[package]]
--- a/docs/TRACING_SETUP.md
+++ b/docs/TRACING_SETUP.md
@@ -1,124 +1,105 @@
-````markdown
 # Distributed Tracing Setup
 
-This guide walks you through setting up **OpenTelemetry distributed tracing** with **Jaeger** or **Uptrace** for the
-CyberFabric framework for testing purposes.
+How to wire the CyberFabric framework to an OpenTelemetry collector (Jaeger,
+Uptrace, Tempo, etc.) and what spans you get out of the box.
 
-## Overview
+## What the framework gives you
 
-The CyberFabric framework includes first-class support for distributed tracing with:
+Without writing any instrumentation code:
 
-- **Automatic trace context extraction** from incoming HTTP requests (W3C Trace Context)
-- **Automatic trace context injection** for outgoing HTTP requests
-- **Centralized configuration** via YAML
-- **HttpClient with OTEL layer** for instrumented HTTP calls
-- **Integration with existing logging**
+- **W3C Trace Context propagation** — incoming `traceparent` header is honored
+  as parent; outgoing HTTP calls made through `modkit_http::HttpClientBuilder`
+  with `.with_otel()` inject it automatically.
+- **Span hierarchy for every REST request:**
+  `http_request` (TraceLayer) → `handler` (route template from `MatchedPath`)
+  → `auth` (jwt validation, principal) → module handler code → `db.insert` /
+  `db.update` / `db.delete` / `db.txn` (SecureConn) and
+  `outgoing_http` → `http.retry` (per attempt) for downstream calls.
+- **Startup and shutdown spans** — `HostRuntime`'s `run_*_phase` methods each
+  get a method-level span (`phase="init" | "db" | "rest" | ...`) with per-module
+  child spans (`module.init`, `module.start`, `db.migrate`, `rest.register`, …).
+- **Real graceful shutdown** — `SdkTracerProvider` and `SdkMeterProvider` are
+  force-flushed and shut down on exit, so the final batch of buffered spans
+  actually reaches the collector.
+- **`tracing_error::ErrorLayer` installed** — error types that opt in via
+  `tracing_error::SpanTrace::capture()` get the current span context for free.
+- **Unified trace-ID extraction** — `modkit::telemetry::current_trace_id()`
+  returns the real W3C trace ID of the active OTEL span (or `None`); use it in
+  `From<DomainError> for Problem` impls and any other place that needs a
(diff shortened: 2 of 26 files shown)


lizreu:
@coderabbitai review

lizreu:
@coderabbitai full review

lizreu:
@coderabbitai full review

lizreu:
@coderabbitai full review

MikeFalcon77 (changes requested):

PR #1606, opened 2026-04-24
ci(clippy): run cargo-hack --each-feature to cover cfg-gated branches

### LINKED ISSUE #1574: Running clippy check with and without feature flags
I noticed in [crypto.rs](https://github.com/cyberfabric/cyberfabric-core/blob/main/libs/modkit/src/bootstrap/crypto.rs) that we currently have two mutually exclusive cfg branches:

`#[cfg(feature = "fips")]`

`#[cfg(not(feature = "fips"))]`

If CI (or local checks) only run clippy with `--features fips` or `--all-features`, clippy never sees the not(fips) branch, and vice versa.

That’s how the `clippy::let-underscore-must-use` warning on line 49 slipped through: it only exists in the non-FIPS build, so a FIPS-only clippy run silently skipped it.

To catch these cases, I suggest we add two clippy passes to CI and makefile:

```bash
cargo clippy --workspace --all-targets --all-features -- -D warnings -D clippy::perf
cargo clippy --workspace --all-targets --no-default-features -- -D warnings -D clippy::perf
```
Alternatively (or additionally), we could introduce something like [cargo-hack](https://crates.io/crates/cargo-hack) and use it in CI to systematically cover different feature combinations.

@MikeFalcon77  is the latest author as I can see

vgprod:
Fixes #1574.
Fixes #1760.
Fixes #1761.

`make clippy` now runs `cargo hack clippy --each-feature`, which performs
one clippy pass per individual feature plus `--no-default-features` and
`--all-features`.

This PR is the minimal "make it pass" change: cargo-hack wiring plus
`allow(...)` / file-level `#[cfg]` suppressions. Two follow-up PRs clean
up specific spots where a proper refactor beats a suppression.

## Stack (open in order)

1. **This PR** — cargo-hack + suppressions
2. #1736 — refactor: crypto provider init uses OnceLock<Result<_>> (draft, depends on this)
3. #1737 — refactor: builder pattern for ModuleCtx db/db_manager fields (draft, depends on #1736)

## Infrastructure
- `make clippy` now runs `cargo hack clippy --each-feature --workspace --all-targets -- -D warnings -D clippy::perf`
- `make setup` installs `cargo-hack`
- CI installs `cargo-hack` alongside `nextest`

## Suppressions / cfg-gates
- `allow(unused_mut)` on a cfg-conditionally-mutated local in `crypto.rs`
- `#[cfg_attr(not(feature = "db"), allow(dead_code))]` on `ModuleCtx`/`ModuleContextBuilder` db fields
- `#[cfg_attr(not(feature = "db"), allow(clippy::unused_async))]` on `for_module`
- `#[cfg_attr(not(feature = "otel"), allow(clippy::unnecessary_wraps))]` on `init_procedure`
- `#[cfg(feature = "db")]` file-level gates on integration tests and a few db-only methods/structs
- `#[cfg(feature = "odata")]` gate on `users-info-sdk`'s `client` module

## Cargo.toml changes (required for the sweep to pass)
- `libs/modkit/Cargo.toml` — `bootstrap` feature now declares its `tracing-subscriber` dep explicitly (was implicitly relying on `otel`)
- `libs/modkit-db/Cargo.toml` — new internal `_any-backend` feature, enabled by any of `sqlite`/`pg`/`mysql`; outbox benches require it (AND-only `required-features` can't express OR)
- Expanded `#![cfg]` on `modkit-db` integration tests to require a backend

## Summary by CodeRabbit

* **Chores**
  * CI and build: install an additional Rust tooling helper and update setup/lint steps to run per-feature checks.
  * Feature config: add a virtual backend feature and require it for certain benchmark targets; adjust a bootstrap feature to include an additional dependency.

* **Refactor**
  * Make many modules and tests compile only when the appropriate feature combinations are enabled, reducing unintended builds.

* **Documentation**
  * Clarify example crate docs and usage with the OData feature and Cargo.toml guidance.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,10 +65,10 @@ jobs:
           rustup show active-toolchain || rustup toolchain install
           rustup show
 
-      - name: Install nextest
+      - name: Install nextest and cargo-hack
         uses: taiki-e/install-action@6ef672efc2b5aabc787a9e94baf4989aa02a97df # v2.70.3
         with:
-          tool: nextest
+          tool: nextest,cargo-hack
 
       # Cache Cargo registry/git + target/ across platforms.
       - name: Cache Cargo/target
--- a/Makefile
+++ b/Makefile
@@ -125,7 +125,9 @@ help:
 ## Install all required development tools
 setup: .setup-stamp
 
-.setup-stamp:
+# Re-run setup whenever the tool list changes (Makefile is the source of truth),
+# so developers who already have .setup-stamp pick up newly-added tools.
+.setup-stamp: Makefile
 	@echo "Installing required development tools..."
 	rustup component add clippy
 	cargo install lychee
@@ -134,6 +136,7 @@ setup: .setup-stamp
 	cargo install cargo-dylint
 	cargo install dylint-link
 	cargo install cargo-fuzz
+	cargo install cargo-hack
 	@if echo "$$OS" | grep -iq windows || [ -n "$$COMSPEC" ]; then \
 		echo "WARNING: kani-verifier and cargo-llvm-cov installation skipped on Windows."; \
 		echo "These tools are not supported on Windows. Use WSL2 or Docker to install instead."; \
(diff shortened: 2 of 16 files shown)


Bechma (commented):
So `cargo hack` runs clippy per feature, that's good, it's a problem we had for some time(building with different feature gates.
However, there's still one open question to me, how to build with a set of feature gates.
Sometimes a crate is intended to be compiled with a specific set of features, not with just one.
Can we override that default behaviour to specify feature sets per crate?

The PR looks good to me, but if we can document that option, it will be very useful in the near future

vgprod:
We could use `--feature-powerset` instead of `--each-feature` but it would check for each combination of features, which would take long time. 

Maybe we could try to add `--group-features` with current setup and see how it works. From my understanding it would treat one group as one feature, but I'm not sure what happens with individual features in that case.
https://crates.io/crates/cargo-hack#options-for-adjusting-the-behavior-of---each-feature-and---feature-powerset

I will test it out

Edit: What would be correct combinations of features?

vgprod:
@Bechma 
I tried using `--group-features` but it requires us to use `--feature-powerset` and we can mimic the behaviour by specifying `--depth`.

To be honest, I'm not sure if I see a benefit of doing it now, since it would also be ok to add another clippy pass that focuses on features we want combined. It would also be clearer to spot and later on we can add groups anytime.

Bechma:
@vgprod that's fair enough, thanks for investigating that, we are on the good track!

Bechma (approved):

MikeFalcon77 (approved):

MikeFalcon77:
Basically I agree but this apporach will increase CI total time significantly.

vgprod:
It will increase it, but it's the only way we could be absolutely sure it compiles correctly with each feature. Smallest alternative would be that we run 3 passes at least


```
cargo clippy --no-default-features
cargo clippy
cargo clippy --all-features
```

Artifizer:
Let's try this approach now and optimize later based on the metrics

Artifizer (approved):

PR #1622, opened 2026-04-26
Modkit OoP requirments and draft design

MikeFalcon77:


docs/arch/modkit/ADR/0002-cpt-cf-adr-auth-edge-only.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Considered Options
+
+* **Option A**: Auth at edge only — gateway validates JWT, OoP modules trust SecurityContext from headers.
+* **Option B**: Auth at every hop — each module validates the JWT independently.
+* **Option C**: Hybrid — gateway validates, modules optionally re-validate (configurable per module).
+
+## Decision Outcome
+
+Chosen option: "Auth at edge only", because it eliminates redundant crypto operations, simplifies module code (no JWKS
```
nonameffh: Can't agree that internal transport must be trustworthy - this allows any pod that can TCP-connect to an OoP module forge an arbitrary `SecurityContext` (any `subject_id`, any tenant) and this can lead  to privilege escalation.

docs/arch/modkit/ADR/0007-cpt-cf-adr-eventual-readiness.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Consequences
+
+* The ModKit OoP runtime (`modkit::bootstrap::oop`) must manage a background task that retries registration with Flight
+  Control (DirectoryService) using exponential backoff. Module code does not participate in this.
+* The ModKit OoP runtime must provide built-in HTTP endpoints for liveness (`/healthz`) and readiness (`/readyz`)probes.
+  These are framework-level, not module-level — the module does not implement them.
+* Liveness returns `200 OK` as soon as the HTTP server is listening (process is alive, not deadlocked).
+* Readiness returns `200 OK` only when all critical dependencies declared in `deps` are resolvable via DirectoryService.
+  Until then, it returns `503 Service Unavailable` with a JSON body listing unresolved deps.
```
nonameffh: This ADR covers only inital phase, what about readiness management on shutdown (for graceful shutdown  of HTTP services you need to set readiness=false before shutting down the service), or dependency failures, graceful degradation? Readiness isn't a simple "all dependencies are up" - it should represent "this instance can really handle a request" (for example, DB connection alive, we're not in shutdown, Redis isn't alive but we can handle that according to the graceful degradation model -> we're ready to handle requests)

PR #1625, opened 2026-04-27
Replace document parsing libraries with Kreuzberg

### LINKED ISSUE #1470: [FileParser] Replace current document parsing libraries with Kreuzberg
We currently use multiple libraries for document parsing:

```toml
# Document parsing libraries
tl = { workspace = true }
pdf-extract = { workspace = true }
calamine = { workspace = true }
pptx-to-md = { workspace = true }
```

This results in fragmented logic across formats (PDF, HTML, Excel, PPTX), most of these libaries lack of functions and do not support specific format features.

Propose replacing these libraries with a single unified solution: **Kreuzberg**.

---

## Rationale

* **Unified pipeline**
  Single API for multiple formats (PDF, DOCX, XLSX, PPTX, HTML, etc.)

* **Consistency**
  Same extraction and normalization logic across all document types

* **Reduced complexity**
  Removes format-specific handling scattered across the codebase

* **Extensibility**
  Built-in support for OCR, chunking, and future RAG-related processing

* **Performance**
  Native Rust implementation (plus C++) with better memory and concurrency characteristics

---

## Scope

* Remove:

  * `tl`
  * `pdf-extract`
  * `calamine`
  * `pptx-to-md`

* Introduce:

  * `kreuzberg`

* Refactor:

  * File parsing module
  * Format detection (delegate to Kreuzberg where possible)
  * Chunking/streaming (need to check)

## Open Questions / Risks
* Verify coverage for all currently supported formats (especially edge cases in PDF and PPTX)
* Evaluate extraction quality vs existing libraries (tables, layout, encoding issues). Run existing tests
* Check binary size and dependency footprint impact
* Benchmark performance on large documents
* Do we need save our IR document format or move to another output format (pure text/MD/IR)

---

## Acceptance Criteria

* All previously supported formats are processed via Kreuzberg
* No regression in extraction quality for existing test corpus

bit8shift:
# Replace document parsing libraries with Kreuzberg

## Problem

The `file-parser` module used four separate libraries for document extraction:

- **tl** — HTML parsing  
- **pdf-extract** — PDF text extraction  
- **calamine** — Excel (xlsx/xls/xlsm/xlsb) parsing  
- **pptx-to-md** — PowerPoint extraction  

This resulted in:
- Fragmented, format-specific logic across four parser files  
- Inconsistent output quality  
- PDF parsing requiring a temp-file workaround (since `pdf-extract` doesn't support in-memory bytes)

---

### Dependencies

- **Removed:** `tl`, `pdf-extract`, `calamine`, `pptx-to-md`  
- **Added:** `kreuzberg 4.9.4` with features:
  - `html`
  - `excel`
  - `office`
  - `pdf`
  - `bundled-pdfium`

### Deleted files

- `html_parser.rs`  
- `pdf_parser.rs`  
- `xlsx_parser.rs`  
- `pptx_parser.rs`  

### New file

- `src/infra/parsers/kreuzberg_parser.rs`  
  - Single `KreuzbergParser` struct handling all four format families via:
    - `kreuzberg::extract_file`
    - `kreuzberg::extract_bytes`

### Modified

- `ir_convert.rs`  
  - Converts Kreuzberg's `ExtractionResult` / `DocumentStructure` into our `ParsedBlock` IR  

- `mod.rs`  
  - Registers `KreuzbergParser`, removes old parsers  

- `module.rs`  
  - Wires `KreuzbergParser` into the parser pipeline  

- `pptx_parser_tests.rs`, `xlsx_parser_tests.rs`  
  - Updated to use `KreuzbergParser`  
  - Assertions relaxed where Kreuzberg output differs from previous libraries  

---

## Behaviour

| Format                         | Previous library                  | Now                                   |
|--------------------------------|----------------------------------|----------------------------------------|
| HTML / HTM                     | `tl`                             | Kreuzberg                              |
| PDF                            | `pdf-extract` (via temp file)    | Kreuzberg (async, in-memory)           |
| XLSX / XLS / XLSM / XLSB       | `calamine`                       | Kreuzberg                              |
| PPTX                           | `pptx-to-md` (via temp file)     | Kreuzberg (async, in-memory)           |
| DOCX                           | `docx-rust`                      | Unchanged                              |
| Images / plain text            | Unchanged                        | Unchanged                              |

Kreuzberg uses `include_document_structure: true`, so the rich `DocumentStructure` path in `ir_convert` is preferred.  
It falls back to plain-text paragraph splitting when structure is unavailable.

---

## Out of scope / follow-ups

- OCR support (`kreuzberg-tesseract` / `kreuzberg-paddle-ocr`) not enabled  
- PDF backend is `bundled-pdfium`; can be switched to:
  - `system-pdfium`
  - `static-pdfium`  
  depending on deployment needs  
- Extraction quality regression testing against the existing test corpus (acceptance criterion from the issue)

Closes #1470 

## Summary by CodeRabbit

* **Refactor**
  * Replaced multiple format-specific parsers with a single, unified parsing backend for consistency.

* **New Features**
  * Unified parser adds broad support for HTML, PDF, Excel, and PowerPoint with improved extraction of headings, paragraphs, lists, tables, images, page breaks, titles, filenames, and language metadata.

* **Tests**
  * Updated unit and e2e tests to target the unified parser and relax strict assumptions where needed.

* **Chores**
  * Updated license allow-list to accept Unicode-DFS-2016 and removed an obsolete proto path reference.


bit8shift:
The `proxy_websocket_connection_header_multi_value failure` is a pre-existing flaky test that has nothing to do with the kreuzberg migration. It's failing in `cf-oagw`, a completely separate crate.

This is a pre-existing flaky test in `cf-oagw` that was already failing on main before this branch was created. This PR only touches `cf-file-parser` and related test/config files.

modules/file-parser/src/infra/parsers/kreuzberg_parser.rs
```diff
@@ -0,0 +1,238 @@
...
+        let mime = content_type
+            .map(str::trim)
+            .filter(|ct| !ct.is_empty() && !ct.eq_ignore_ascii_case("application/octet-stream"))
+            .or_else(|| {
+                filename_hint
+                    .and_then(|name| Path::new(name).extension())
+                    .and_then(|ext| ext.to_str())
+                    .and_then(|ext| Self::mime_for_ext(ext))
+            })
+            .ok_or_else(|| {
```
MikeFalcon77: **HIGH**

The wrapper rejects unknown content_type and `application/octet-stream` before calling `extract_bytes`, but `extract_bytes` itself runs `mime::detect_mime_type_from_bytes` for `application/octet-stream` (see `kreuzberg::core::extractor::bytes`). Forcing the early `Cannot determine MIME type` error removes a working detection path for unlabeled uploads.

It also regresses behavior from the old `XlsxParser`, which sniffed magic bytes (`OLE_MAGIC` / `ZIP_MAGIC`) when the filename hint was missing.

When `content_type` is empty/unknown and the extension lookup fails, pass `"application/octet-stream"` through to `extract_bytes` and let Kreuzberg sniff. Only error out if Kreuzberg also fails to identify the format.

MikeFalcon77:
### Streaming / memory under concurrent uploads

`parse_bytes` holds the entire upload in RAM until extraction finishes. Under concurrent cloud traffic this is an OOM source — peak per-request RSS is `file_size + sizeof(ExtractionResult)`, multiplied by in-flight requests. The current `max_file_size_mb` is enforced at `service.rs:180`, *after* `handlers.rs:217` has already collected the full body via `field.bytes().await`, so it doesn't bound RAM at all.

**No streaming in Kreuzberg.** Verified against 4.7.4 (this PR) and current 4.9.4. Public API is `extract_file(impl AsRef<Path>, ...)`, `extract_bytes(&[u8], ...)`, and `batch_extract_*` (parallel across files, not within one). Extractors do iterate page-by-page internally (`pdf/text.rs`, `extraction/excel.rs`), but everything is aggregated into a single `ExtractionResult` before it returns. The `DocumentExtractor` plugin trait still returns `Result<InternalDocument>` whole; its "streaming" doc comment refers to how a plugin reads its input, not its return shape. Real streaming would need an upstream API change (or bypassing Kreuzberg, which defeats this PR).

**What can be done now, no upstream changes:**

1. **Spool to a tempfile, call `extract_file` instead of `extract_bytes`.** pdfium memory-maps PDF pages and releases them as it iterates; calamine reads sheet ranges on demand from the open workbook. Peak per-request RSS drops from `file_size + sizeof(ExtractionResult)` to ≈ `sizeof(ExtractionResult)`. Concretely in `kreuzberg_parser.rs::parse_bytes`: write `bytes` to `tempfile::NamedTempFile`, delegate to `parse_local_path`, drop the temp on return. Better: stream the multipart field straight to disk in `handlers.rs` so `Bytes` never materialises — replace `field.bytes().await` with a `field.chunk()` loop into `tokio::fs::File`.

2. **Bounded concurrency at the service layer.** Weighted `tokio::sync::Semaphore`, permit cost ≈ expected peak working set per format (rough: PDF ~4× file_size, Excel/PPTX ~2×, HTML ~1×). Sum of permits ≈ available RAM budget. Without this, even disk-spooled uploads OOM the pod at N concurrent M-byte requests. Surface `max_concurrent_parses` and `max_in_flight_bytes` next to `max_file_size_mb` in `FileParserConfig`.

3. **Reject oversize uploads at the HTTP boundary, before buffering.** `axum::extract::DefaultBodyLimit::max(...)` on the multipart routes plus a `Content-Length` check in the handler. Today a hostile or buggy client can fill RAM with multi-GB before the `service.rs:180` check fires.

(1) is the cheap order-of-magnitude RAM win. (2) and (3) are required regardless for any cloud deployment. None of these block the migration itself, but the OOM exposure was lower with the old `pdf-extract` / `pptx-to-md` path because those used temp files internally — this PR removes that mitigation without replacing it.

bit8shift:
additional comments, putting it here for tracking purposes 
from @MikeFalcon77 : Could we support OCR (feature-gated) for PDF with pictures?

and from @ainetx : could you please remove ignoring of specs here https://github.com/cyberfabric/cyberfabric-core/blob/main/.cypilot/config/artifacts.toml#L87-L89
and update and fix the specs - according to the current state of the module - https://github.com/cyberfabric/cyberfabric-core/tree/main/modules/file-parser/docs

bit8shift:
@MikeFalcon77 
Kreuzberg already has OCR support built in
```
[features]
ocr = ["kreuzberg/ocr"]  # pulls in kreuzberg-tesseract + image + fast_image_resize
```
`ExtractionConfig` already has pub `ocr: Option<OcrConfig>` , it can be set it in `KreuzbergParser::config()` when the feature is enabled
```
fn config() -> ExtractionConfig {
    ExtractionConfig {
        include_document_structure: true,
        #[cfg(feature = "ocr")]
        ocr: Some(OcrConfig {
            backend: "tesseract".to_string(),
            language: "eng".to_string(),
            ..Default::default()
        }),
        ..Default::default()
    }
}
```
No changes to `ir_convert.rs`, OCR output comes through the same `result.content` / `result.document` paths already

Contraints:
- Requires Tesseract system library installed (libtesseract-dev) — not zero-dependency
- kreuzberg/ocr is Tesseract-based; kreuzberg/paddle-ocr is an alternative but brings heavier deps
- Still pinned at =4.7.4 due to the license boundary — OCR support at that version is confirmed present

MikeFalcon77 (approved):

bit8shift:
```
Error: docs/api/api.json does not match the spec generated from code.

  To fix: run 'make openapi' locally and commit the updated docs/api/api.json.

Error: Process completed with exit code 1.
```

this is a false positive, I ran `make openapi` the file generated is exactly the same as the one in commit.

```
git diff --stat origin/main -- docs/api/api.json
```
it is identical to `origin/main` after regeneration

Artifizer (changes requested):
Our original idea of the File Parser was to have a gateway and plugins, that define supported file types and priority. So the Kreuzberg must be just a plugin, because tomorrow we can have others (including Tika, LibreOffice, IBM's one, etc)

Can you please change the design to become a plugin?

bit8shift:
> Our original idea of the File Parser was to have a gateway and plugins, that define supported file types and priority. So the Kreuzberg must be just a plugin, because tomorrow we can have others (including Tika, LibreOffice, IBM's one, etc)
> 
> Can you please change the design to become a plugin?

@Artifizer 
`KreuzbergParser` is already a plugin, 
`module.rs` currently registers 5 plugins:
`PlainTextParser → KreuzbergParser → DocxParser → ImageParser → StubParser`
 
`KreuzbergParser` only claims its extensions (pdf, html, xlsx, pptx, etc.). The service routes to whichever plugin's supported_extensions() matches first.

also, updated `DESIGN.md` and `PRD.md` to reflect this `gateway + plugin` design, and also updated to include all other plugins instead of just Kreuzberg.

Artifizer (approved):
As agreed, let's have separate changes in https://github.com/cyberfabric/cyberfabric-core/issues/1762, https://github.com/cyberfabric/cyberfabric-core/issues/1763, https://github.com/cyberfabric/cyberfabric-core/issues/1764

PR #1746, opened 2026-04-28
feat(resource-group): add tenant_id to GroupFilterField whitelist

### LINKED ISSUE #1626: Add `hierarchy/tenant_id` to `GroupFilterField` whitelist
## Problem

A parent tenant must be able to list resource groups owned by a specific child tenant — e.g. AM's `soft_delete` precondition checks "does this child own any RG?" before allowing deletion.

Today the only way to scope `list_groups` to one child tenant is to anchor on a tenant-RG row whose `id == tenant_id` (via `get_group_descendants`). We are deliberately **not** creating such a row on `create_tenant`: it introduces a second lifecycle that can drift from AM's `tenants` table — a manual `delete_group` or a partial-failure in the create saga silently breaks consistency.

The clean alternative is to filter `list_groups` by `hierarchy/tenant_id`, but that field is not in the OData whitelist (`resource-group-sdk/src/odata/groups.rs`) — only `type`, `hierarchy/parent_id`, `id`, `name` are. Calling `list_groups(ctx, $filter=hierarchy/tenant_id eq <child>, $top=1)` is rejected by validation.

Without this filter AM cannot narrow to a specific child: parent's `AccessScope` already covers all descendants (PRD §4), so an unfiltered `list_groups($top=1)` returns non-empty whenever any sibling has resources, over-blocking `soft_delete`.

## Required change

1. Add `HierarchyTenantId` to `GroupFilterField`: `name = "hierarchy/tenant_id"`, `kind = Uuid`, ops `eq`/`ne`/`in`.
2. Map it to the `resource_group.tenant_id` column in the SQL repo backing `list_groups`.
3. Update the doc-block at the top of `odata/groups.rs`.

diffora:
Allows callers to scope list_groups to a single tenant via $filter=tenant_id eq <uuid>. Required by AM's soft_delete precondition (does this child own any RG?) per #1626.

Closes #1626

## Summary by CodeRabbit

* **New Features**
  * Tenant ID filtering is now supported for groups in OData queries, letting users filter and retrieve groups by associated tenant.

* **Tests**
  * Added/updated tests to validate tenant ID field mapping, OData naming, and UUID type classification.

* **Documentation**
  * API docs, design, and feature docs updated to list tenant_id as a supported OData filter.

### DIFF
--- a/modules/system/resource-group/docs/DECOMPOSITION.md
+++ b/modules/system/resource-group/docs/DECOMPOSITION.md
@@ -178,7 +178,7 @@ The Resource Group DESIGN is decomposed into seven features organized around the
   - Hierarchy queries: ancestors/descendants ordered by depth via indexed closure table lookups
   - Hierarchy depth endpoint: `GET /groups/{group_id}/hierarchy` returning `ResourceGroupWithDepth` with relative depth (positive = descendants, negative = ancestors, 0 = self) and OData filtering on `hierarchy/depth`
   - Query profile enforcement: configurable `max_depth`/`max_width` on writes, no truncation on reads for already-existing data, deterministic `DepthLimitExceeded`/`WidthLimitExceeded` errors
-  - Group REST endpoints: CRUD under `/api/resource-group/v1/groups` with OData `$filter` on `type`, `hierarchy/parent_id`, `id`, `name`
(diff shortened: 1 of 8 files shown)


ffedoroff:
approved from my side

Artifizer (approved):

